PR #100, opened 2019-05-24
Tao 8229 cache warmer enhancements

SergiiTao:


src/Command/Cache/DoctrineResultCacheWarmerCommand.php
```diff
@@ -119,47 +176,85 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                ->where('l.id IN (:lineItemIds)')
+                ->setParameter('lineItemIds', $this->lineItemIds);
+        }
+
+        $result = $queryBuilder
             ->getQuery()
-            ->getSingleScalarResult();
+            ->getOneOrNullResult();
+
+        return (int)$result['number_of_users'];
```
Alroniks: getOneOrNullResult may return null, in this case `$result['number_of_users']` seems unsafe.

uncleempty: Indeed, it will throw a notice, if null is returned.

PR #101, opened 2019-07-09
Feature/tao 6655/introduce user group

Skyld:
Ticket: https://oat-sa.atlassian.net/browse/TAO-8655
Follow up ticket after this: https://oat-sa.atlassian.net/browse/TAO-8654

### Added
- `groupId` property to `User` entity for logical grouping of users.

### Changed
- Upgraded to Symfony version `4.3`

### DIFF
--- a/config/doctrine/User.orm.yml
+++ b/config/doctrine/User.orm.yml
@@ -5,35 +5,32 @@ App\Entity\User:
     id:
         id:
             type: integer
-            scale: 0
             length: null
             nullable: false
-            precision: 0
             id: true
             generator:
                 strategy: AUTO
     fields:
         username:
             type: string
-            scale: 0
             length: 255
             unique: true
             nullable: false
-            precision: 0
         password:
             type: string
-            scale: 0
             length: 255
             unique: false
             nullable: false
-            precision: 0
         roles:
             type: json_array
-            scale: 0
             length: null
             unique: false
             nullable: false
-            precision: 0
+        groupId:
+            type: string
+            length: 255
+            unique: false
+            nullable: true
     oneToMany:
         assignments:
             targetEntity: App\Entity\Assignment
--- a/config/doctrine/User.orm.yml
+++ b/config/doctrine/User.orm.yml
@@ -31,6 +31,7 @@ App\Entity\User:
             length: 255
             unique: false
             nullable: true
+            column: group_id
     oneToMany:
         assignments:
             targetEntity: App\Entity\Assignment
--- a/src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
+++ b/src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
@@ -139,7 +139,7 @@ abstract class AbstractBulkUsersAssignmentsCommand extends Command
     /**
      * @throws LogicException
      */
-    protected function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
+    private function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
     {
         if (!$output instanceof ConsoleOutputInterface) {
             throw new LogicException(
--- a/src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
+++ b/src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
@@ -139,7 +139,7 @@ abstract class AbstractBulkUsersAssignmentsCommand extends Command
     /**
      * @throws LogicException
      */
-    private function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
+    protected function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
     {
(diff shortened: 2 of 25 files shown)


config/doctrine/User.orm.yml
```diff
@@ -5,35 +5,32 @@ App\Entity\User:
...
             type: json_array
-            scale: 0
             length: null
             unique: false
             nullable: false
-            precision: 0
+        groupId:
+            type: string
+            length: 255
+            unique: false
```
uncleempty: Just to check, as we still don't have full info on groups:
should it really be nullable?

src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
```diff
@@ -139,7 +139,7 @@ protected function processOperationCollection(BulkOperationCollection $bulkOpera
     /**
      * @throws LogicException
      */
-    protected function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
+    private function ensureConsoleOutput(OutputInterface $output): ConsoleOutputInterface
```
uncleempty: Should it really be a private method, since the class itself is abstract?

config/doctrine/User.orm.yml
Skyld: Yes, because for example Invalsi doesn't use groups. We'll implement a load balancing strategy (in the next pr) on top of the group, but other than that, it's not a must have column.

uncleempty: Aw, yeah, indeed, forgot that roster is not client-specific.

src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
Skyld: It's not a method meant to be overwritten by inherited classes.

Skyld:
> Line 67 in `src/EventSubscriber/RequestIdGeneratorSubscriber.php`:
> should we really call getter method when we already have requestId in subscriber?

I would say it doesn't matter -> so not REALLY :)

src/Command/Bulk/AbstractBulkUsersAssignmentsCommand.php
uncleempty: Yep, but what I meant is:
a) `execute` method is inherited from `Command` class, and left protected, that's ok - classes extending `AbstractBulkUsersAssignmentsCommand` will either use its own implementation of `execute` method, or will just "call" the parent one;
BUT
b) since `AbstractBulkUsersAssignmentsCommand` is an abstract class - making `ensureConsoleOutput` a private method will also make it a "non-callable" from any inheriting class, and it can't be called from the class itself, since class is 1) abstract; 2) non-static


So, if some class will extend `AbstractBulkUsersAssignmentsCommand` and use smth like `parent::execute` in its own `execute` method -> code will fail

uncleempty: So the question itself:
either we should reconsider methods' visibility or completely remove it?

Skyld: Ah ok I see your point. In theory yes, there is a possibility for error, so I reverted it to protected.

tests/Integration/Ingester/Ingester/UserIngesterTest.php
```diff
@@ -125,6 +132,29 @@ public function testIngestWithValidSource(): void
...
+
+        $usersInGroup1 = $this->getRepository(User::class)->findBy(['groupId' => 'group_1']);
+        $this->assertCount(3, $usersInGroup1);
+
+        $usersInGroup2 = $this->getRepository(User::class)->findBy(['groupId' => 'group_2']);
+        $this->assertCount(4, $usersInGroup2);
+
+        $usersInGroup3 = $this->getRepository(User::class)->findBy(['groupId' => 'group_3']);
+        $this->assertCount(5, $usersInGroup3);
+    }
```
SergiiTao: $usersInGroup4 = $this->getRepository(User::class)->findBy(['groupId' => 'group_4']);

$usersInGroup4 === ?

config/doctrine/User.orm.yml
```diff
@@ -5,35 +5,32 @@ App\Entity\User:
...
             nullable: false
-            precision: 0
         roles:
             type: json_array
-            scale: 0
             length: null
             unique: false
             nullable: false
-            precision: 0
+        groupId:
```
SergiiTao: do we have a naming convention about field names in DB?
usually, field names should be lowercased and `_` separated

tests/Integration/Ingester/Ingester/UserIngesterTest.php
Skyld: It's ok as we discussed

config/doctrine/User.orm.yml
Skyld: Nice catch, fixed

PR #102, opened 2019-07-10
Implemented user group id based load balancing strategy

Skyld:
https://oat-sa.atlassian.net/browse/TAO-8654

Introduced LTI load balancing interface, implemented `User` group id based LTI load balancing strategy.


.env
```diff
@@ -38,7 +38,9 @@ ASSIGNMENT_STATE_INTERVAL_THRESHOLD=P1D
 ###> LTI ###
 LTI_ENABLE_INSTANCES_LOAD_BALANCER=false
 LTI_LAUNCH_PRESENTATION_LOCALE=en-EN
-LTI_LAUNCH_PRESENTATION_RETURN_URL='http://localhost:8000/index.html'
+LTI_LAUNCH_PRESENTATION_RETURN_URL=http://localhost:8000/index.html
+# Possible values for LTI Instance load balancer strategy: username or userGroupId
```
Alroniks: `username` and `usergroup` looks more nicely

src/Lti/LoadBalancer/UserGroupIdLtiInstanceLoadBalancer.php
```diff
@@ -0,0 +1,40 @@
...
+
+class UserGroupIdLtiInstanceLoadBalancer extends AbstractLtiInstanceLoadBalancer
+{
+    /**
+     * @throws IndeterminableLtiInstanceUrlException
+     */
+    public function getLtiInstanceUrl(User $user): string
+    {
+        if (!$user->hasGroupId()) {
+            throw new IndeterminableLtiInstanceUrlException(
```
Alroniks: What to do if the user does not have group id? Just exit with errors?

.env
Alroniks: Also, does it make sense to provide `.env.example` instead of changing the real .env file?

src/Lti/LoadBalancer/LtiInstanceLoadBalancerFactory.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function __construct(array $ltiInstances)
+    {
+        $this->ltiInstances = $ltiInstances;
+    }
+
+    /**
+     * @throws LogicException
+     */
+    public function __invoke(string $loadBalancerStrategy): LtiInstanceLoadBalancerInterface
```
SergiiTao: why magic instead of commonly used `create`?

src/Lti/Exception/IndeterminableLtiInstanceUrlException.php
```diff
@@ -0,0 +1,26 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ *  Copyright (c) 2019 (original work) Open Assessment Technologies S.A.
+ */
+
+namespace App\Lti\Exception;
+
+class IndeterminableLtiInstanceUrlException extends \RuntimeException
```
SergiiTao: simplify FQN?

.env
Skyld: About usergroup: I like the idea, however it's not camelCase. username is a word on it's own, but usergroup is invalid, userGroup would be the appropriate format :(
About the .env file: It contains default, but working implementation, but I agree that it could be improved. I'll create a ticket for that 👍

src/Lti/LoadBalancer/UserGroupIdLtiInstanceLoadBalancer.php
Skyld: Yes. IMHO It's better to exit instead of forwarding every user to the same instance due to missing group id.

src/Lti/LoadBalancer/LtiInstanceLoadBalancerFactory.php
Skyld: Invokable factories for services were introduced in Symfony 4.3. It's way better in my opinion than hardcoding method names in the container.

src/Lti/Exception/IndeterminableLtiInstanceUrlException.php
Skyld: I don't know, for me this is okay, it's not too long.

src/Lti/LoadBalancer/AbstractLtiInstanceLoadBalancer.php
```diff
@@ -32,7 +33,7 @@ public function __construct(array $ltiInstances)
         $this->ltiInstances = $ltiInstances;
     }
 
-    public function getLoadBalancedLtiInstanceUrl(string $value): string
+    protected function getLoadBalancedLtiInstanceUrl(string $value): string
```
SergiiTao: I'd go with keeping it a separated class 
it will increase flexibility in case we need to implement different algorithms 
IMHO algorithms should be unittested

src/Lti/Exception/IndeterminableLtiInstanceUrlException.php
SergiiTao: for me also, for PHPStorm no :)

src/Lti/LoadBalancer/AbstractLtiInstanceLoadBalancer.php
Skyld: Yes you are right, but the fact we have an injected load balancer here is a hack. We've a ticket somewhere in the backlog to extract the load balancer from here to a separate application, so I would not spend time now on it (YAGNI) if you don't mind.

src/Lti/Exception/IndeterminableLtiInstanceUrlException.php
Skyld: Fixed

src/Lti/LoadBalancer/LtiInstanceLoadBalancerFactory.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        switch ($loadBalancerStrategy) {
```
uncleempty: But wouldn't it be better to check like `in_array` then using `switch/case` with `default` action?
p.s.
I know that "tastes differ", just curious about it

src/Lti/LoadBalancer/UserGroupIdLtiInstanceLoadBalancer.php
uncleempty: And just to check from my side:
since we have the `groupId` field nullable in [here](https://github.com/oat-sa/simple-roster/pull/101/files#diff-078a85fb273a48b46623a07626dab15bR33) should we really throw an exception and exit? Even if this is a `UserGroupIdLoadBalancer`?

src/Lti/LoadBalancer/LtiInstanceLoadBalancerFactory.php
Skyld: What is the difference?

src/Lti/LoadBalancer/UserGroupIdLtiInstanceLoadBalancer.php
Skyld: As I said above, it's a requirement to have group id in order to use this strategy. Otherwise it would forward all the users to the same instance, and I think it's not what we want :)

src/Lti/LoadBalancer/LtiInstanceLoadBalancerFactory.php
uncleempty: No real difference :) It's more about tastes
p.s.
Plus, just for me `in_array` is a bit easier to read. But, again, it's just a question of taste

src/Lti/LoadBalancer/UserGroupIdLtiInstanceLoadBalancer.php
uncleempty: Yep, agree. Just one moment: Depp has some users (smth in `admin_users.csv` files) that really doesn't relate to any group -> as I remember it's for an internal "testing" of teachers.
p.s.
Ofc, we can workaround it with just assigning some "generated" group for such users :)

Skyld: We'll use the school id as group id if you remember. Can you confirm that all students belong to a school?

uncleempty: It was in their files. But I don't really remember if it was them or someone other having such accounts (for internal "teachers training") with a group that wasn't existent in other files -> so roster ingestion failed because of using unknow group (it was just a stub for a UAI field, that came from their side).
Smth like that, sorry, that's the only thing I remember.

Skyld: I think we can address these issues while working on the CSV adaptations. We'll need a school id for every user, otherwise the proctoring won't work

uncleempty: Agreed. Thanks for the answers :+1:

PR #108, opened 2019-07-23
Feature/tao 8729 possibility to use group id as lti context

Skyld:
### Added
- ~~Added `force` parameter to `NativeUserIngesterCommand`~~
- Added possibility to determine the `contextId` LTI request parameter based on LTI load balancing strategy.

### Fixed
- ~~Fixed fatal error during composer install caused by https://bugs.php.net/bug.php?id=76980 on php `7.2.20`.~~

Investigation: https://oat-sa.atlassian.net/wiki/spaces/EXP/pages/105677222/Implementation+plans+-+DEPP+-+Implementing+LTI+Director+for+proctoring+tool+in+Principal+Portal
Ticket: https://oat-sa.atlassian.net/browse/TAO-8728


src/Entity/User.php
```diff
@@ -124,6 +124,11 @@ public function addAssignment(Assignment $assignment): self
         return $this;
     }
 
+    public function hasLastAssignment(): bool
```
uncleempty: I'm not really sure in method name.
Is `has...` method should really be named like `hasLastSomething`, since it's only checking if `something[]` is not empty?

Skyld: You are right, I forgot to update the method name. It will be `hasAssignments()`, good catch 👍

PR #113, opened 2019-09-30
Code quality fixes

Skyld:
Quality of life improvements for next major release (1.5.0)

Follow up ticket created for next sprint: https://oat-sa.atlassian.net/browse/TAO-9319


src/Lti/Service/GetUserAssignmentLtiRequestService.php
```diff
@@ -137,7 +140,7 @@ private function getAssignmentLtiLink(Assignment $assignment): string
         return sprintf(
             '%s/%s',
             $link,
-            base64_encode(json_encode(['delivery' => $assignment->getLineItem()->getUri()]))
+            base64_encode((string)json_encode(['delivery' => $assignment->getLineItem()->getUri()]))
```
SergiiTao: in case of failing of `json_encode` we will end up with `base64_encode('')`
so if we're expecting an error here i'd checked for === false or if we are running under php 7.3 it is possible to add `JSON_THROW_ON_ERROR` option.
Btw. sometimes on errors `json_encode` returns nothing. it is not documented, but it happens. So i'd even check for `empty`

Skyld: Nice catch, I'll create a ticket about this (to specify `JSON_THROW_ON_ERROR` option after upgrading to php 7.3)

PR #116, opened 2020-01-28
Added dependency security checker into composer

Skyld:
- Added dependency security checker into composer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,5 +76,5 @@
 ## 1.1.0 - 2019-03-19
 
 ### Added
-- Added possibility to specify charset for regular and native ingester.
+- Added possibility to specify charset for regular and native ingesters.
 - Added possibility to perform integrity checks on ingester dry runs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
     - Running PHPUnit code coverage checker
     - Running mutation tests with Infection
     - Running static code analysis with PHPStan
+- Added package dependency security checker into composer. 
 
 ### Changed
 - Changed password encoding algorithm from hardcoded Argon2i to [automatic](https://symfony.com/blog/new-in-symfony-4-3-native-password-encoder).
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -75,5 +76,5 @@
 ## 1.1.0 - 2019-03-19
 
 ### Added
-- Added possibility to specify charset for regular and native ingesters.
+- Added possibility to specify charset for regular and native ingester.
```
luisfn: Not sure why not use the previous plural, since it's related to the 2 types (regular/native).

config/bundles.php
```diff
@@ -28,12 +10,12 @@
...
     Symfony\Bundle\MonologBundle\MonologBundle::class => ['all' => true],
-    Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle::class => ['all' => true],
     Doctrine\Bundle\DoctrineBundle\DoctrineBundle::class => ['all' => true],
     Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle::class => ['all' => true],
     Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle::class => ['dev' => true, 'test' => true],
     Nelmio\Alice\Bridge\Symfony\NelmioAliceBundle::class => ['dev' => true, 'test' => true],
     Fidry\AliceDataFixtures\Bridge\Symfony\FidryAliceDataFixturesBundle::class => ['dev' => true, 'test' => true],
     Hautelook\AliceBundle\HautelookAliceBundle::class => ['dev' => true, 'test' => true],
     Nelmio\CorsBundle\NelmioCorsBundle::class => ['all' => true],
+    Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle::class => ['all' => true],
```
luisfn: As curiosity, why this order change?

CHANGELOG.md
Skyld: Mmmmm, maybe you are right, reverted.

config/bundles.php
Skyld: This file automatically updated by Symfony. I removed the doctrine cache bundle from terminal then I re-added it (I was testing something) and it appended it at the end of the list. The order of the bundles here is not relevant.

PR #117, opened 2020-02-24
Changed ORM mapping driver from YAML to XML

Skyld:
- Changed ORM mapping driver from YAML to XML
- Fixed some deprecation errors related to functional tests (KernelBrowser usage)
- Cleaned up database fixture traits


tests/Functional/Action/HealthCheck/HealthCheckActionTest.php
```diff
@@ -22,14 +22,14 @@
 
 namespace App\Tests\Functional\Action\HealthCheck;
 
-use App\Tests\Traits\DatabaseTrait;
+use App\Tests\Traits\DatabaseTestingTrait;
```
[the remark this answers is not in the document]
Skyld: No, it was not used, nice catch :)

config/doctrine/Assignment.orm.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+    <entity repository-class="App\Repository\AssignmentRepository" name="App\Entity\Assignment" table="assignments">
+        <id name="id" type="integer" column="id">
+            <generator strategy="AUTO"/>
```
uncleempty: Shouldn't the strategy be set to `SEQUENCE` or it was left as `AUTO` for "cross-vendor" stuff?

config/doctrine/Assignment.orm.xml
```diff
@@ -0,0 +1,29 @@
...
+            <sequence-generator sequence-name="assignments_id_seq" allocation-size="1" initial-value="1"/>
+        </id>
+        <field name="state" type="string" column="state" length="255" precision="0" scale="0" nullable="false"/>
+        <field name="updatedAt" type="datetime" column="updated_at" nullable="true"/>
+        <many-to-one field="user" target-entity="App\Entity\User" inversed-by="assignments" fetch="LAZY">
```
uncleempty: Just a side question:
since we still have one user per assignment should the strategy really be left as `LAZY`?

tests/Functional/Action/Lti/UpdateLtiOutcomeActionTest.php
```diff
@@ -28,27 +28,36 @@
 use App\Repository\InfrastructureRepository;
 use App\Security\OAuth\OAuthContext;
 use App\Security\OAuth\OAuthSigner;
-use App\Tests\Traits\DatabaseFixturesTrait;
+use App\Tests\Traits\DatabaseTestingTrait;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
```
uncleempty: Why should we have it here, only for line 40? Or am I missing smth?

tests/Functional/Action/Security/LoginActionTest.php
```diff
@@ -22,20 +22,32 @@
 
 namespace App\Tests\Functional\Action\Security;
 
-use App\Tests\Traits\DatabaseFixturesTrait;
+use App\Tests\Traits\DatabaseTestingTrait;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
```
uncleempty: Same here, it looks like unneeded use statement

tests/Integration/Security/Provider/UserProviderTest.php
```diff
@@ -25,7 +25,7 @@
 use App\Entity\User;
 use App\Repository\UserRepository;
```
uncleempty: Same here: is it needed?

config/doctrine/Assignment.orm.xml
Skyld: It must be set to AUTO, because of cross vendor functionality yes, plus in test environment we use SQLite, which does not have sequences.

config/doctrine/Assignment.orm.xml
Skyld: It's not true that we have only one. For example if you use the lambda assignment manager, you can have any number of assignments per user.

tests/Functional/Action/Lti/UpdateLtiOutcomeActionTest.php
Skyld: Sorry I don't get the question

tests/Functional/Action/Security/LoginActionTest.php
Skyld: Why it in unneded?

PR #136, opened 2020-07-02
Feature/tdp 57 add max attempts to line items

Babacooll:
Ticket : https://oat-sa.atlassian.net/browse/TDP-57

_This pull request adds a new property maxAttempts to the LineItem entity in order to be able to have multiple assignments per user for a single line item._

* Added a new maxAttempts (int) to the LineItem entity & its normalized version
* Fixed the LineItemIngested to handle the new maxAttempts property
* Fixed Open API
* Fixed fixtures
* Fixed tests

✔️ All tests run with 100% coverage
✔️ PHPStan runs with level _max_
✔️ Infection runs successfully with multiple threads (99%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## 1.5.0 - TODO: To be released
+## 1.6.0 - To be released
+
+### Added
+- Added new property `maxAttempts` to the `LineItem` entity.
+- Added support for ingestion of `maxAttempts` on the `LineItemIngester`.
+
+## 1.5.0 - 2020-06-17
 
 ### Added
 - Added dedicated `docker` application environment for development purposes.
--- a/config/doctrine/LineItem.orm.xml
+++ b/config/doctrine/LineItem.orm.xml
@@ -18,6 +18,10 @@
                 <join-column name="infrastructure_id" referenced-column-name="id"/>
             </join-columns>
         </many-to-one>
-        <field name="maxAttempts" type="integer" column="max_attempts" precision="0" scale="0" nullable="true"/>
+        <field name="maxAttempts" type="integer" column="max_attempts" precision="0" scale="0" nullable="true">
+            <options>
+                <option name="default">0</option>
+            </options>
+        </field>
     </entity>
 </doctrine-mapping>
--- a/config/doctrine/LineItem.orm.xml
+++ b/config/doctrine/LineItem.orm.xml
@@ -18,5 +18,6 @@
                 <join-column name="infrastructure_id" referenced-column-name="id"/>
             </join-columns>
         </many-to-one>
+        <field name="maxAttempts" type="integer" column="max_attempts" precision="0" scale="0" nullable="true"/>
     </entity>
 </doctrine-mapping>
--- a/src/Entity/LineItem.php
+++ b/src/Entity/LineItem.php
@@ -48,6 +48,9 @@ class LineItem implements JsonSerializable, EntityInterface
     /** @var Infrastructure */
     private $infrastructure;
 
+    /** @var int */
+    private $maxAttempts;
+
     public function getId(): ?int
     {
         return $this->id;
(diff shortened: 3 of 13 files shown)


Babacooll:
* LineItem.orm.xml fixed
* Tests improved
* Changelog fixed

But I'm not sure about the "Update documentation (Ingestion doc, devops doc)" as there is no mention there of any field related to LineItem ingestion.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
-## 1.5.0 - TODO: To be released
+## 1.6.0 - To be released
```
uncleempty, line 3: not a blocker or smth, but would be changed in future: I will re-check, looks like it was not backmerged by me to develop after 1.5.0 release

config/doctrine/LineItem.orm.xml
```diff
@@ -18,5 +18,6 @@
                 <join-column name="infrastructure_id" referenced-column-name="id"/>
             </join-columns>
         </many-to-one>
+        <field name="maxAttempts" type="integer" column="max_attempts" precision="0" scale="0" nullable="true"/>
```
uncleempty, line 21: How do you think, maybe it would be better to have default = 0?
Smth like:
```xml

        <field name="maxAttempts" type="integer" column="max_attempts" nullable="false">
            <options>
                <option name="default">0</option>
            </options>
        </field>
```

src/Entity/LineItem.php
```diff
@@ -48,6 +48,9 @@ class LineItem implements JsonSerializable, EntityInterface
     /** @var Infrastructure */
     private $infrastructure;
 
+    /** @var int */
+    private $maxAttempts;
```
uncleempty, line 52: Same proposal here:
maybe smth like
```php
    /** @var int */
    private $maxAttempts = 0;
```

src/Entity/LineItem.php
```diff
@@ -134,6 +137,18 @@ public function isAvailableForDate(DateTimeInterface $date): bool
         return $this->startAt <= $date && $this->endAt >= $date;
     }
 
+    public function getMaxAttempts(): ?int
```
uncleempty, line 140: And in case if previous proposals would be accepted (and as soon as it doesn't makes much sense to have null for infinite attempts), please, change the return type hint of this method

tests/Resources/Ingester/Invalid/line-items.csv
```diff
@@ -1,5 +1,5 @@
...
-http://taoplatform.loc/delivery_1.rdf,label1,gra13_ita_1,infra_1,1546682400,1546713000
-http://taoplatform.loc/delivery_2.rdf,label2,gra13_ita_1,infra_2,1546682400,1546713000
-http://taoplatform.loc/delivery_3.rdf,label3,gra13_ita_1,infra_3,,1546713000
-http://taoplatform.loc/delivery_4.rdf,label4,gra13_ita_1,infra_4,1546682400,
\ No newline at end of file
+uri,label,slug,infrastructure,startTimestamp,endTimestamp,maxAttempts
+http://taoplatform.loc/delivery_1.rdf,label1,gra13_ita_1,infra_1,1546682400,1546713000,0
+http://taoplatform.loc/delivery_2.rdf,label2,gra13_ita_1,infra_2,1546682400,1546713000,0
+http://taoplatform.loc/delivery_3.rdf,label3,gra13_ita_1,infra_3,,1546713000,0
+http://taoplatform.loc/delivery_4.rdf,label4,gra13_ita_1,infra_4,1546682400,0,
```
uncleempty, line 5: Last line should be:
```csv
http://taoplatform.loc/delivery_4.rdf,label4,gra13_ita_1,infra_4,1546682400,,0
```

src/Entity/LineItem.php
```diff
@@ -137,7 +137,7 @@ public function isAvailableForDate(DateTimeInterface $date): bool
         return $this->startAt <= $date && $this->endAt >= $date;
     }
 
-    public function getMaxAttempts(): int
+    public function getMaxAttempts(): ?int
```
luisfn, line 140: Maybe it could return a default value (0,1) if not set? Not sure it makes more sense do it here or in a later point when using it. What do you think?

PR #142, opened 2020-07-06
Feature/tdp 58 add attempt to assignment

Babacooll:
* Added new property _attemptsCount_ to _Assignment_ entity
* Upon user ingestion assignments is created with 0 as attempt count
* Attempt counter is increased upon LTI launch
* LTI outcome call sets the assignment status to READY if no maximum attempts or current attempt count is lower than max attempts
* Garbage collector command sets the assignment status to READY if no maximum attempts or current attempt count is lower than max attempts
* New property _attemptsCount_ is be returned in every API response

✔️ Tests running
✔️ PHPStan with max level
✔️ Infection 99% MSI

### DIFF
--- a/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
+++ b/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
@@ -58,7 +58,8 @@ class AssignmentGarbageCollectorCommand extends Command
         AssignmentRepository $assignmentRepository,
         LoggerInterface $logger,
         string $cleanUpInterval
-    ) {
+    )
+    {
         parent::__construct(self::NAME);
 
         $this->assignmentRepository = $assignmentRepository;
@@ -141,7 +142,8 @@ class AssignmentGarbageCollectorCommand extends Command
             /** @var Assignment $assignment */
             $assignmentCount = $stuckAssignments->getIterator()->count();
             foreach ($stuckAssignments as $assignment) {
-                $assignment->setState(Assignment::STATE_COMPLETED);
+                $assignment->complete();
+
                 if (!$isDryRun) {
                     $this->assignmentRepository->persist($assignment);
                 }
--- a/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
+++ b/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
@@ -58,8 +58,7 @@ class AssignmentGarbageCollectorCommand extends Command
         AssignmentRepository $assignmentRepository,
         LoggerInterface $logger,
         string $cleanUpInterval
-    )
-    {
+    ) {
         parent::__construct(self::NAME);
 
         $this->assignmentRepository = $assignmentRepository;
--- a/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
+++ b/src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
@@ -109,9 +109,8 @@ class AssignmentGarbageCollectorCommand extends Command
 
             $successMessage = $numberOfCollectedAssignments !== 0
                 ? sprintf(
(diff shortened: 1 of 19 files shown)


src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
```diff
@@ -108,9 +108,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
             $successMessage = $numberOfCollectedAssignments !== 0
                 ? sprintf(
-                    "Total of '%s' stuck assignments were successfully marked as '%s'.",
-                    $numberOfCollectedAssignments,
-                    Assignment::STATE_COMPLETED
+                    "Total of '%s' stuck assignments were successfully collected.",
+                    $numberOfCollectedAssignments
                 )
                 : 'Nothing to update.';
```
uncleempty, line 114: Just a proposal: maybe it'd be better for readability to split ternary operator?

tests/Unit/Entity/AssignmentTest.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
```
uncleempty, line 3: please move it after the copyright block :)

src/Entity/Assignment.php
```diff
@@ -134,12 +137,45 @@ public function isCancellable(): bool
...
+    {
+        $this->attemptsCount = $attemptsCount;
+
+        return $this;
+    }
+
+    public function incrementAttemptsCount(): self
+    {
+        $this->attemptsCount++;
+
```
uncleempty, line 155: Shouldn't we have a check somewhere upon calling this? Mean that it won't be the case that `attemptsCount` would be incremented to the value > than maxAttampts.
Or there is already such a check, I'm just blind? :upside_down_face:

tests/Unit/Entity/AssignmentTest.php
Babacooll: It's always before the copyright block in the whole project. Are you sure ?

uncleempty: Good question. The point is that, as I see in backend channel - we had an agreement to have `declare` block after the copyright statement (messages from 03/26 and 03/27). Maybe it's not a good point of time to change this for the whole project, but we can stick to this for the new files, no? :upside_down_face:

src/Entity/Assignment.php
Babacooll: You're right it was missing, fixed it :) !

src/Command/GarbageCollector/AssignmentGarbageCollectorCommand.php
Babacooll: Can't figure out a better way to display it unfortunately, and I don't want to shorten the message for code style purposes

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
```diff
@@ -158,9 +180,10 @@ public function testItReturnsLtiLinkAndUpdatedAssignmentStateToStartedWithUserna
         /** @var Assignment $assignment */
         $assignment = $this->getRepository(Assignment::class)->find(1);
         $this->assertEquals(Assignment::STATE_STARTED, $assignment->getState());
+        $this->assertEquals(2, $assignment->getAttemptsCount());
     }
 
-    public function testItReturnsLtiLinkAndUpdatedAssignmentStateToStartedWithUserGroupIdLoadBalancerStrategy(): void
+    public function testItReturnsLtiLinkAndUpdatedAssignmentStateToStartedAndIncrementsAttemptsCountWithUserGroupIdLoadBalancerStrategy(): void
```
uncleempty, line 186: We are aiming to beat Java with the class/method names length :joy:

PR #143, opened 2020-07-06
Feature/tdp 61 modify listuserassignmentsaction

Babacooll:
The _/api/v1/assignments_ endpoint now returns **all available user assignments** without filtering on their availability.

✔️ Tests passing
✔️ PHPStan passing
✔️ Infection passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Changed garbage collection state update to `ready` if `Assignment` has additional attempts available.
 - Increment the `attemptsCount` on upon LTI Launch if state is not started.
 - Set default value of `0` for `attemptsCount` for new assignments during user ingestion.
-- The `/api/v1//assignments` endpoint now returns all users assignments (available or not).
+- The `/api/v1/assignments` endpoint now returns all users assignments (available or not).
 
 ## 1.5.0 - 2020-06-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Changed garbage collection state update to `ready` if `Assignment` has additional attempts available.
 - Increment the `attemptsCount` on upon LTI Launch.
 - Set default value of `0` for `attemptsCount` for new assignments during user ingestion.
+- The `/api/v1//assignments` endpoint now returns all users assignments (available or not).
 
 ## 1.5.0 - 2020-06-17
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Changed garbage collection state update to `ready` if `Assignment` has additional attempts available.
 - Increment the `attemptsCount` on upon LTI Launch if state is not started.
 - Set default value of `0` for `attemptsCount` for new assignments during user ingestion.
+- The `/api/v1//assignments` endpoint now returns all users assignments (available or not).
```
uncleempty, line 15: please fix the typo (duplicated `/`)

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
```diff
@@ -112,6 +112,32 @@ public function testItReturns409IfAssignmentDoesNotHaveASuitableState(): void
...
+
+        $this->logInAs($user, $this->kernelBrowser);
+
+        $this->kernelBrowser->request('GET', '/api/v1/assignments/1/lti-link');
+
+        $this->assertEquals(Response::HTTP_NOT_FOUND, $this->kernelBrowser->getResponse()->getStatusCode());
+
+        $decodedResponse = json_decode($this->kernelBrowser->getResponse()->getContent(), true);
+        $this->assertEquals(
+            "Assignment id '1' not found for user 'user1'.",
```
uncleempty, line 134: The usual question, which is though not related to test (and possibly could be fixed in another PR/task):
should we really return 404s when user is trying to launch assignment which is yet unavailable, but he/she sees assignment in list?

Babacooll: I do agree with you. But I think that it's not only the 404 that is wrong, the message is as well. The assignment is found, but we cannot generate an LTI link, so looks more like conflict to me. I'll fix it in another task that I'll create as it touches many endpoints.

CHANGELOG.md
Babacooll: Good catch, fixed.

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
Babacooll: https://oat-sa.atlassian.net/browse/TDP-63

Skyld: I confirm this is indeed an issue, it should throw a 409 exception if we try to retrieve LTI link for existing, but unavailable assignments.

PR #161, opened 2020-10-14
Native ingestion and cache warmup performance improvements

Skyld:
### Changed
- Switched from offset based pagination to cursor based in `DoctrineResultCacheWarmerCommand` for better performance.
- Switched from ORM to native queries in `NativeUserIngesterCommand` for better performance.

### Fixed
- Fixed bug in `DoctrineResultCacheWarmerCommand` where lack of order by clause caused wrong pagination with PostgreSQL.

### DIFF
--- a/src/Command/Cache/DoctrineResultCacheWarmerCommand.php
+++ b/src/Command/Cache/DoctrineResultCacheWarmerCommand.php
@@ -204,9 +204,6 @@ class DoctrineResultCacheWarmerCommand extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $offset = 0;
-        $numberOfWarmedUpCacheEntries = 0;
-
         $this->symfonyStyle->note('Calculating total number of entries to warm up...');
 
         $numberOfTotalUsers = $this->getNumberOfTotalUsers();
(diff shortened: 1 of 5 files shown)


src/Command/Cache/DoctrineResultCacheWarmerCommand.php
```diff
@@ -224,13 +221,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        $numberOfWarmedUpCacheEntries = 0;
         do {
             $iterateResult = $this
-                ->getFindAllUsernameQuery($offset)
+                ->getFindAllUsernameQuery($lastUserId)
                 ->iterate();
 
             foreach ($iterateResult as $row) {
-                $this->warmUpResultCacheForUserName(current($row)['username']);
+                $username = current($row)['username'];
```
luisfn, line 232: Are these current() calls really needed?

uncleempty: As I see - they are. But, what I would propose is to move the `curent()` call to an additional variable:
```
$currentRow = current($row)
```
and then just
```
$username = $currentRow['username'];
```

uncleempty (approved):
👍 
Awesome job, man!
p.s
Just a small thingy to change, IMO, but already approved ;)

bugalot (approved):
Sounds good to me, but I only "flew over it". The experiments in production seem to me good, let's make this happen in reality by releasing this.

PR #166, opened 2020-11-10
[RO-17] Add support for lti1p3 link generation

luisfn:
Ticket: https://oat-sa.atlassian.net/browse/RO-17

### Added
- Added environment variable to control LTI version on link generation
- Added specific request factories, one per LTI version
- Added `oat-sa/bundle-lti1p3` bundle for LTI 1.3 link creation
- Added [Features documentation](docs/features.md)


.env
```diff
@@ -43,8 +43,12 @@ ASSIGNMENT_STATE_INTERVAL_THRESHOLD=P1D
 LTI_LAUNCH_PRESENTATION_LOCALE=en-EN
 LTI_LAUNCH_PRESENTATION_RETURN_URL=http://localhost:8000/index.html
 LTI_OUTCOME_XML_NAMESPACE=http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0
+# Possible values for LTI Version: LTI-1p0 or 1.3.0
+LTI_VERSION=LTI-1p0
```
uncleempty, line 47: There is naming inconsistency. Please use either `LTI-1p0` / `LTI-1p3` or `1.0.0` / `1.3.0`

.env.docker
```diff
@@ -42,9 +42,13 @@ ASSIGNMENT_STATE_INTERVAL_THRESHOLD=P1D
 ###> LTI ###
 LTI_LAUNCH_PRESENTATION_LOCALE=en-EN
 LTI_LAUNCH_PRESENTATION_RETURN_URL=http://localhost:8000/index.html
+# Possible values for LTI Version: LTI-1p0 or 1.3.0
```
uncleempty, line 45: same thing about naming inconsistency

.env.test
```diff
@@ -36,9 +36,13 @@ CACHE_TTL_LTI_INSTANCES=3600
 ###> LTI ###
 LTI_LAUNCH_PRESENTATION_LOCALE=en-EN
 LTI_LAUNCH_PRESENTATION_RETURN_URL='http://example.com/index.html'
+# Possible values for LTI Version: LTI-1p0 or 1.3.0
```
uncleempty, line 39: and here

src/Lti/Exception/InvalidGroupException.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Lti\Exception;
+
+use RuntimeException;
+
+class InvalidGroupException extends RuntimeException
+{
+
+}
```
uncleempty, line 30: I've seen the usage of this class, but is there any sense of having an empty class, with a message provided from the place exception is thrown? Meaning that `InvalidGroupException` doesn't anyhow extend RuntimeException. Because for now it doesn't differ from throwing `RuntimeException` with custom message.
Maybe it would make more sense to move the message inside the `InvalidGroupException` class as the default one?

src/Lti/Exception/InvalidLtiVersionException.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Lti\Exception;
+
+use RuntimeException;
+
+class InvalidLtiVersionException extends RuntimeException
+{
+
+}
```
uncleempty, line 30: I've seen the usage of this class, but is there any sense of having an empty class, with a message provided from the place exception is thrown? Meaning that `InvalidLtiVersionException` doesn't anyhow extend RuntimeException. Because for now it doesn't differ from throwing `RuntimeException` with custom message.
Maybe it would make more sense to move the message inside the `InvalidLtiVersionException` class as the default one?

src/Lti/Exception/RegistrationNotFoundException.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Lti\Exception;
+
+use LogicException;
+
+class RegistrationNotFoundException extends LogicException
+{
+
+}
```
uncleempty, line 30: I've seen the usage of this class, but is there any sense of having an empty class, with a message provided from the place exception is thrown? Meaning that `RegistrationNotFoundException` doesn't anyhow extend RuntimeException. Because for now it doesn't differ from throwing `RuntimeException` with custom message.
Maybe it would make more sense to move the message inside the `RegistrationNotFoundException` class as the default one?

src/Lti/Factory/Lti1p1RequestFactory.php
```diff
@@ -0,0 +1,157 @@
...
+    private function getLtiLaunchLink(LtiInstance $ltiInstance, LineItem $lineItem): string
+    {
+        return sprintf(
+            '%s/ltiDeliveryProvider/DeliveryTool/launch/%s',
+            rtrim($ltiInstance->getLtiLink(), '/'),
+            base64_encode(
+                json_encode(
+                    ['delivery' => $lineItem->getUri()],
+                    JSON_THROW_ON_ERROR,
+                    512
```
uncleempty, line 131: Just a check, why do we need to set `depth = 512`, since `512` is the default value?

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,86 @@
...
+            (string) $loginHint,
+            null,
+            [
+                LtiRequest::LTI_ROLE
+            ],
+            [
+                new ContextClaim(
+                    (string)$assignment->getId(),
+                    [],
+                    $assignment->getLineItem()->getSlug(),
```
uncleempty, line 78: Just a matter of taste, but `$assignment->getLineItem()` call is used 4 times, wouldn't it be easier to read/understand when `$assignment->getLineItem()` is extracted to separate variable?

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,86 @@
...
+        RegistrationRepositoryInterface $repository,
+        LtiResourceLinkLaunchRequestBuilder $builder
+    ) {
+        $this->repository = $repository;
+        $this->builder = $builder;
+    }
+
+    public function create(Assignment $assignment): LtiRequest
+    {
+        $registrationId = 'demo';
```
uncleempty, line 52: `registrationId` would always be a `demo` one? If no - why it's a "method constant". If yes, why using additional variable?

src/Lti/Factory/LtiRequestFactory.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function __construct(
+        string $ltiVersion,
+        Lti1p1RequestFactory $lti1p1RequestFactory,
+        Lti1p3RequestFactory $lti1p3RequestFactory
+    ) {
+        $this->ltiVersion = $ltiVersion;
+        $this->lti1p1RequestFactory = $lti1p1RequestFactory;
+        $this->lti1p3RequestFactory = $lti1p3RequestFactory;
+    }
```
uncleempty, line 47: Just not sure if it can't be done the other way, but why this "factory" should have both services injected in any case, not depending on which is actually used? Shouldn't it create them depending on env variable?

src/Lti/Factory/LtiRequestFactory.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Lti\Factory;
+
+use OAT\SimpleRoster\Lti\Exception\InvalidLtiVersionException;
+use OAT\SimpleRoster\Lti\Request\LtiRequest;
+
+class LtiRequestFactory
```
uncleempty, line 28: It's a bit strange to have `LtiRequestFactory` class which doesn't implement `LtiRequestFactoryInterface`.
p.s.
Though I understand why, it's really may be unclear for anyone else.

src/Lti/Request/LtiRequest.php
```diff
@@ -27,18 +27,23 @@
 class LtiRequest implements JsonSerializable
 {
     public const LTI_MESSAGE_TYPE = 'basic-lti-launch-request';
-    public const LTI_VERSION = 'LTI-1p0';
+    public const LTI_VERSION_1P0 = 'LTI-1p0';
+    public const LTI_VERSION_1P3 = '1.3.0';
```
uncleempty, line 31: Again, a question about naming inconsistency between LTI version implementations. Please consider "unification"

src/Lti/Service/GetUserAssignmentLtiRequestService.php
```diff
@@ -22,102 +22,29 @@
...
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Symfony\Component\Routing\RouterInterface;
 
 class GetUserAssignmentLtiRequestService
 {
-    /** @var OAuthSigner */
-    private $signer;
+    /** @var LtiRequestFactoryInterface */
+    private $ltiRequestFactory;
```
uncleempty, line 33: Again the same problem with naming consistency. Though `LtiRequestFactory` (the one without version) is not implementing the `LtiRequestFactoryInterface` - looking on the names I would expect a possibility to have it under this variable

src/Security/Lti/LoginHintValidator.php
```diff
@@ -0,0 +1,75 @@
...
+                break;
+            }
+        }
+
+        if (!$lineItemFound) {
+            throw new LineItemNotFoundException(
+                sprintf(
+                    'Line Item with slug %s not found for username %s',
+                    $loginHintDto->getSlug(),
+                    $loginHintDto->getUsername()
```
uncleempty, line 68: Again, just a matter of taste, but I would extract those two into additional variables, since they're used more than once

src/Security/Lti/LoginHintValidator.php
```diff
@@ -0,0 +1,75 @@
...
+    /** @var UserRepository */
+    private $userRepository;
+
+    public function __construct(LoginHintExtractor $loginHintExtractor, UserRepository $userRepository)
+    {
+        $this->loginHintExtractor = $loginHintExtractor;
+        $this->userRepository = $userRepository;
+    }
+
+    public function validate(string $loginHint): User
```
uncleempty, line 45: Just a matter of taste/readability:
why validator method is returning smth except `true` / `false`? IMO - breaks SRP

src/Security/Lti/OidcUserAuthenticator.php
```diff
@@ -0,0 +1,51 @@
...
+    private $loginHintValidator;
+
+    public function __construct(LoginHintValidator $loginHintValidator)
+    {
+        $this->loginHintValidator = $loginHintValidator;
+    }
+
+    public function authenticate(string $loginHint): UserAuthenticationResultInterface
+    {
+        $user = $this->loginHintValidator->validate($loginHint);
```
uncleempty, line 44: Addition to one of the previous comments:
it's a bit strange to expect validator to return User object, because of SRP

tests/Traits/KeyPairTestingTrait.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+trait KeyPairTestingTrait
+{
+    protected function generateKeyPairs(): void
+    {
+        mkdir('tests/Resources/secrets');
+
+        shell_exec('openssl genrsa -out tests/Resources/secrets/private.key 2048 2> /dev/null');
+        shell_exec('openssl rsa -in tests/Resources/secrets/private.key -outform PEM' .
+            ' -pubout -out tests/Resources/secrets/public.key 2> /dev/null');
```
uncleempty, line 38: Even if it's just a trait for testing purposes - I'm not sure that `openssl` lib is in the default packages list for PHP, better to test this case (especially, since, this app is used not only under docker/linux, but also can be used on win/mac)

tests/Unit/Security/Lti/LoginHintValidatorTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->loginHintExtractor
+            ->expects($this->once())
+            ->method('extract')
+            ->with($loginHint)
+            ->willReturn($loginHintDto);
+
+        $result = $this->subject->validate($loginHint);
+
+        $this->assertSame($result, $user);
+    }
```
uncleempty, line 86: Again, it's a bit inconsistent (though may be the only choice) to have `testLoginHintIsValid` which is calling `validate` method and not asserting returned value is boolean true, but objects are the same

uncleempty (changes requested):
Good job :+1: 
Though some changes are still to be added

.env
luisfn: I would prefer it as well and in fact started with LTI-1p0 / LTI-1p3.
The problem here is that these are the names defined by IMS standards. If I change it here, it would probably change the value on the responses as well, unless I use different constants for this setting and for the responses. 
We need some agreement here. What do you think @Skyld @andreluizmachado

src/Lti/Exception/InvalidGroupException.php
luisfn: Its a good point, but it's how it was being done for the other exceptions as well. As I can see, we usually don't catch treat exceptions (they are bubbling up) but if we do, It would help to differ on what when wrong, even with messages being defined by the "thrower".

tests/Traits/KeyPairTestingTrait.php
luisfn: For mac we are covered for sure, but for windows I have no idea how to proceed (or i we should care)

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,86 @@
...
+            throw new RegistrationNotFoundException(sprintf('Registration %s not found.', $registrationId));
+        }
+
+        $loginHint = new LoginHintDto(
+            $assignment->getUser()->getUsername(),
+            $assignment->getUser()->getGroupId(),
+            $assignment->getLineItem()->getSlug()
+        );
+
+        $message = $this->builder->buildLtiResourceLinkLaunchRequest(
```
andreluizmachado, line 66: Don't we have parameters missing?
because we're sending all of this on 1.1, but not here:
```
		             'lti_message_type' => LtiRequest::LTI_MESSAGE_TYPE,
             'lti_version' => LtiRequest::LTI_VERSION_1P0,
             'context_id' => $this->loadBalancer->getLtiRequestContextId($assignment),
             'roles' => LtiRequest::LTI_ROLE,
             'user_id' => $assignment->getUser()->getUsername(),
             'lis_person_name_full' => $assignment->getUser()->getUsername(),
             'resource_link_id' => $assignment->getId(),
             'lis_outcome_service_url' => $this->router->generate(
                 'updateLtiOutcome',
                 [],
                 UrlGeneratorInterface::ABSOLUTE_URL
             ),
             'lis_result_sourcedid' => $assignment->getId(),
             'launch_presentation_return_url' => $this->ltiLaunchPresentationReturnUrl,
             'launch_presentation_locale' => $this->ltiLaunchPresentationLocale,
         ];
```

This document is doing the comparative:
https://www.imsglobal.org/spec/lti/v1p3/migr
Maybe it'd be good just double check it. :)

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
```diff
@@ -256,7 +264,7 @@ public function testItReturnsLtiLinkAndUpdatedAssignmentAndAttemptsCountWithUser
                     'oauth_timestamp' => (string)Carbon::getTestNow()->getTimestamp(),
                     'oauth_version' => OAuthContext::VERSION_1_0,
                     'lti_message_type' => LtiRequest::LTI_MESSAGE_TYPE,
-                    'lti_version' => LtiRequest::LTI_VERSION,
+                    'lti_version' => LtiRequest::LTI_VERSION_1P0,
```
andreluizmachado, line 267: I'd suggest using a data provider to test both cases 1.1 and 1.3

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
```diff
@@ -203,7 +207,7 @@ public function testItReturnsLtiLinkAndUpdatedAssignmentAndAttemptsCountWithUser
...
```
andreluizmachado, line 210: I'd suggest using a data provider to test both cases 1.1 and 1.3

andreluizmachado (changes requested):
Good Job @luisfn. Once you have time, please check my suggestions.

src/Security/Lti/LoginHintValidator.php
luisfn: I think it's normal to do checking like these (not really returning boolean) but agree that its is a little misleading. Will rework it a little.

src/Security/Lti/OidcUserAuthenticator.php
luisfn: Thanks, will update it

tests/Unit/Security/Lti/LoginHintValidatorTest.php
luisfn: Removed the validator, checking login hint consistency directly on OidcUserAuthenticator

src/Lti/Extractor/LoginHintExtractor.php
```diff
@@ -0,0 +1,57 @@
...
+            throw new LogicException('Missing username on login hint.');
+        }
+
+        if (empty($matches['slug'])) {
+            throw new LogicException('Missing slug on login hint.');
+        }
+
+        return new LoginHintDto(
+            $matches['username'],
+            $matches['groupId'],
```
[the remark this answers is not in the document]
luisfn: It will return an empty string for this case. That's why it is not validated on the DTO

tests/Unit/Security/Lti/LoginHintValidatorTest.php
luisfn: We don't have the validator anymore

src/DataTransferObject/LoginHintDto.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /** @var string */
+    private $groupId;
+
+    /** @var string */
+    private $slug;
+
+    public function __construct(string $username, string $groupId, string $slug)
+    {
+        if (empty($username)) {
```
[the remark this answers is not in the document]
luisfn: When trying to extract it from the loginHInt, if absent it will be an empty string

src/Lti/Extractor/LoginHintExtractor.php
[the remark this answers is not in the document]
luisfn: It can be empty string

src/Lti/Factory/Lti1p3RequestFactory.php
luisfn: Not sure what different did you spotted, can you please be more specific? I believe for 1.3 all data goes on the "claims" section.

http://www.imsglobal.org/spec/lti/v1p3/#message-claims

In this case, the lti1p3-bundle already create the necessary claims. We are just adding the context claim

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
luisfn: We definitely could, but in this case they are built in a completely different way, so I don't see a easy way to use providers

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
luisfn: We definitely could, but in this case they are built in a completely different way, so I don't see a easy way to use providers

.env
Skyld: My preference would be to use the format `1.x.x`. If the IMS spec is not consistent about the version formats, let's ignore it wherever we can (in configurations / API responses / code).

src/DataTransferObject/LoginHintDto.php
```diff
@@ -0,0 +1,72 @@
...
+    /** @var string */
+    private $username;
+
+    /** @var string */
+    private $groupId;
+
+    /** @var string */
+    private $slug;
+
+    public function __construct(string $username, string $groupId, string $slug)
```
Skyld, line 38: I like the concept of having a VO for login hint but I'm afraid you've picked the wrong parameters here.

What do we want to know from a login hint?
- Who is the user (= `username` ✔️ )
- Which assignment are we trying to launch ( = assignment ➖ )

The line item is not specific enough information, because a user can have multiple assignments for the same line item. With your current implementation we would generate the same login hint for all of them.

Also group id is redundant information here, the username identifies the user already, so it can be removed.

src/Lti/Factory/Lti1p1RequestFactory.php
```diff
@@ -0,0 +1,157 @@
...
+use OAT\SimpleRoster\Generator\NonceGenerator;
+use OAT\SimpleRoster\Lti\LoadBalancer\LtiInstanceLoadBalancerInterface;
+use OAT\SimpleRoster\Lti\Request\LtiRequest;
+use OAT\SimpleRoster\Security\OAuth\OAuthContext;
+use OAT\SimpleRoster\Security\OAuth\OAuthSigner;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+class Lti1p1RequestFactory implements LtiRequestFactoryInterface
```
Skyld, line 40: Can you confirm that this class did not change, the logic was just extracted from `OAT\SimpleRoster\Lti\ServiceGetUserAssignmentLtiRequestService`?

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,90 @@
...
+use OAT\Library\Lti1p3Core\Resource\LtiResourceLink\LtiResourceLink;
+use OAT\SimpleRoster\DataTransferObject\LoginHintDto;
+use OAT\SimpleRoster\Entity\Assignment;
+use OAT\SimpleRoster\Lti\Exception\RegistrationNotFoundException;
+use OAT\SimpleRoster\Lti\Request\LtiRequest;
+
+class Lti1p3RequestFactory implements LtiRequestFactoryInterface
+{
+    /** @var RegistrationRepositoryInterface */
+    private $repository;
```
Skyld, line 37: $registrationRepository

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,90 @@
...
+
+        if (!$registration) {
+            throw new RegistrationNotFoundException(sprintf('Registration %s not found.', $this->ltiRegistrationId));
+        }
+
+        $loginHint = new LoginHintDto(
+            (string) $assignment->getUser()->getUsername(),
+            (string) $assignment->getUser()->getGroupId(),
+            $assignment->getLineItem()->getSlug()
+        );
```
Skyld, line 68: Basically you could take the `Assignment` as constructor parameter and extract all necessary information from it.

Skyld: Also could you please remove all the spaces between the typecasts and variables. Like: `(string)$assignment->getUser()->getUsername()`

src/Lti/Factory/Lti1p3RequestFactory.php
Skyld: There are no parameters in case of LTI 1.3 launch, that's why they are not here, they are needed only for LTI 1.1

src/Lti/Factory/LtiRequestFactory.php
Skyld: Yes, this is conceptually wrong. Basically in any case you are instantiating one of the factories for no reason. They must be instantiated here. Please check for example here: OAT\SimpleRoster\Lti\LoadBalancer\LtiInstanceLoadBalancerFactory`

src/Lti/Factory/LtiRequestFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function __invoke(): LtiRequestFactoryInterface
+    {
+        switch ($this->ltiVersion) {
+            case LtiRequest::LTI_VERSION_1P0:
+                return $this->lti1p1RequestFactory;
+            case LtiRequest::LTI_VERSION_1P3:
+                return $this->lti1p3RequestFactory;
+            default:
+                throw new InvalidLtiVersionException('Invalid LTI Version specified: ' . $this->ltiVersion);
```
Skyld, line 60: You should throw `LogicException` here instead since this is a program logic error.

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,90 @@
...
+        $message = $this->builder->buildLtiResourceLinkLaunchRequest(
+            $resourceLink,
+            $registration,
+            (string) $loginHint,
+            null,
+            [
+                LtiRequest::LTI_ROLE
+            ],
+            [
+                new ContextClaim(
```
Skyld, line 79: Could you please check with Jonathan how can we add support for the following parameters:

- `launch_presentation_return_url`
- `launch_presentation_locale`

We offer support for these in LTI 1.1, would be nice to be consistent and send these information here as well. I assume there also claims for them, or we need to define custom ones.

src/Lti/Factory/LtiRequestFactoryInterface.php
```diff
@@ -0,0 +1,36 @@
...
+namespace OAT\SimpleRoster\Lti\Factory;
+
+use OAT\SimpleRoster\Entity\Assignment;
+use OAT\SimpleRoster\Lti\Request\LtiRequest;
+
+/**
+ * LtiRequestFactoryInterface
+ *
+ * @package OAT\SimpleRoster\Lti\Factory
+ */
```
Skyld, line 32: You can remove this docblock

src/Security/Lti/OidcUserAuthenticator.php
```diff
@@ -0,0 +1,90 @@
...
+use OAT\SimpleRoster\DataTransferObject\LoginHintDto;
+use OAT\SimpleRoster\Entity\User;
+use OAT\SimpleRoster\Exception\LineItemNotFoundException;
+use OAT\SimpleRoster\Lti\Exception\InvalidGroupException;
+use OAT\SimpleRoster\Lti\Extractor\LoginHintExtractor;
+use OAT\SimpleRoster\Repository\UserRepository;
+
+class OidcUserAuthenticator implements UserAuthenticatorInterface
+{
+    public const UNDEFINED_USERNAME = 'Undefined Username';
```
Skyld, line 38: The constant is not used anywhere, it can be removed.

src/Security/Lti/OidcUserAuthenticator.php
```diff
@@ -0,0 +1,90 @@
...
+    /** @var UserRepository */
+    private $userRepository;
+
+    public function __construct(LoginHintExtractor $loginHintExtractor, UserRepository $userRepository)
+    {
+        $this->loginHintExtractor = $loginHintExtractor;
+        $this->userRepository = $userRepository;
+    }
+
+    public function authenticate(string $loginHint): UserAuthenticationResultInterface
```
Skyld, line 52: You should wrap the logic within a try catch block to make sure you are not bubbling up any exceptions. In case of error the interface expects to send a `success = false` within the result without user identity.

Skyld: Also we need to add logs here about the result of the authentication

src/Security/Lti/OidcUserAuthenticator.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $this->checkLoginHintConsistency($user, $loginHintDto);
+
+        return new UserAuthenticationResult(
+            true,
+            new UserIdentity((string) $user->getUsername())
+        );
+    }
+
+    private function checkLoginHintConsistency(User $user, LoginHintDto $loginHintDto): void
```
Skyld, line 65: This method is supposed to change based on my comments on `LoginHintDto` so I won't comment here.

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
```diff
@@ -394,5 +408,64 @@ static function (array $record) use ($assignment, $ltiRequestInResponse) {
...
+            'lti_deployment_id=1',
+            urldecode($responseData['ltiLink'])
+        );
+        self::assertStringContainsString(
+            'client_id=demo',
+            urldecode($responseData['ltiLink'])
+        );
+
+        self::assertSame(LtiRequest::LTI_VERSION_1P3, $responseData['ltiVersion']);
+        self::assertSame([], $responseData['ltiParams']);
```
Skyld, line 460: This is not really nice. Let's decode the **full** response and assert on it like in the other test cases above please.

tests/Traits/KeyPairTestingTrait.php
```diff
@@ -0,0 +1,47 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Tests\Traits;
+
+/**
+ * KeyPairTestingTrait
+ *
+ * @package OAT\SimpleRoster\Tests\Traits
+ */
+trait KeyPairTestingTrait
```
Skyld, line 30: I know that we discussed about this yesterday, but I believe we can remove this trait. Let's pregenerate the keys in config/secrets/test` folder and commit them into the repository.

tests/Unit/DataTransferObject/LoginHintDtoTest.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->expectExceptionMessage('Slug cannot be empty');
+
+        new LoginHintDto('username', 'group', '');
+    }
+
+    public function testLoginHintDtoCreation(): void
+    {
+        $loginHintDto = new LoginHintDto('username', 'groupId', 'slug');
+
+        $this->assertSame('username', $loginHintDto->getUsername());
```
Skyld, line 51: please invoke all assertion methods statically (everywhere)

tests/Unit/Lti/Extractor/LoginHintExtractorTest.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function testShouldExtractDataWhenLoginHintIsWellFormed(
+        string $loginHint,
+        string $user,
+        string $groupdId,
+        string $slug
+    ): void {
+        $loginHintDto = $this->subject->extract($loginHint);
+
+        $this->assertSame($user, $loginHintDto->getUsername());
```
Skyld, line 52: please invoke all assertion methods statically (everywhere)

tests/Unit/Lti/Factory/Lti1p1RequestFactoryTest.php
```diff
@@ -0,0 +1,191 @@
...
+            $this->loadBalancer,
+            $this->ltiLaunchPresentationReturnUrl,
+            $this->ltiLaunchPresentationLocale,
+        );
+    }
+
+    /**
+     * @dataProvider provideValidLineItemAssignmentCombinations
+     */
+    /**
```
Skyld, line 75: Please remove

tests/Unit/Lti/Factory/Lti1p1RequestFactoryTest.php
```diff
@@ -0,0 +1,191 @@
...
+    /** @var LtiInstanceLoadBalancerInterface|MockObject */
+    private $loadBalancer;
+
+    /** @var string */
+    private $ltiLaunchPresentationReturnUrl;
+
+    /** @var string */
+    private $ltiLaunchPresentationLocale;
+
+
```
Skyld, line 53: Please remove extra line

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
```diff
@@ -0,0 +1,179 @@
...
+    ): void {
+        Carbon::setTestNow(Carbon::createFromDate(2020, 1, 1));
+
+        $assignment = $this->createAssignmentMock($maxAttempts, $attemptsCount, $assignmentStatus);
+
+        $this->createRegistrationMock();
+
+        $message = $this->createMock(LtiMessageInterface::class);
+        $message
+            ->expects($this->once())
```
Skyld, line 76: `self::once()` (please fix it everywhere)

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
```diff
@@ -0,0 +1,179 @@
...
+            ->setUsername('testUsername')
+            ->setGroupId('groupId');
+
+        $assignment = $this->createPartialMock(Assignment::class, ['getId']);
+
+        $assignment
+            ->method('getId')
+            ->willReturn(5);
+
+        $assignment
```
Skyld, line 161: You can return here

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
```diff
@@ -0,0 +1,179 @@
...
+            ->setLineItem($lineItem)
+            ->setUser($user)
+            ->setAttemptsCount($attemptsCount)
+            ->setState($assignmentStatus);
+
+        return $assignment;
+    }
+
+    private function createRegistrationMock(): void
```
Skyld, line 170: This method does not really add value because, also it does more than what it says. You can remove and set mock expectations in your test case

Skyld (changes requested):

.env
luisfn: Changed to 1.1.1 and 1.3.0

src/Security/Lti/LoginHintValidator.php
luisfn: Validation now happens on the OidcUserAuthenticator

config/packages/lti1p3.yaml
```diff
@@ -0,0 +1,36 @@
...
+        platformKey:
+            key_set_name: "platformSet"
+            public_key: "file://%kernel.project_dir%/config/secrets/public.key"
+            private_key: "file://%kernel.project_dir%/config/secrets/private.key"
+            private_key_passphrase: null
+        demoKey:
+            key_set_name: "demoSet"
+            public_key: "file://%kernel.project_dir%/config/secrets/demo.key"
+            private_key: null
+            private_key_passphrase: null
```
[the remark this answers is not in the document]
luisfn: You are correct, we don't need the second one.

tests/Unit/Lti/Factory/Lti1p1RequestFactoryTest.php
luisfn: Removed

tests/Unit/Lti/Factory/Lti1p1RequestFactoryTest.php
luisfn: Removed, thanks

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
luisfn: Thanks

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
luisfn: I use it on bot tests

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
luisfn: Can you tell me why? And  should do it to everything? `self::createMock()`, `self::assertSomething()`?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - Added `usernames` and `line-item-slugs` input options to [roster:doctrine-result-cache:warmup](docs/cli/doctrine-result-cache-warmer-command.md) command.
 - Added default value (`local`) to `source` argument of [roster:ingest](docs/cli/ingester-command.md) command.
 - Added default value (`local`) to `source` argument of [roster:native-ingest:user](docs/cli/native-user-ingester-command.md) command.
+- Added environment variable to control LTI version on link generation
+- Added specific request factories, one per LTI version
```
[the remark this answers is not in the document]
luisfn: Added! Thanks!

tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php
luisfn: There is no way to do it since we are generating different keys. I think it will also generate different values even with the same key pair,  that's why I'm checking parts of the link

tests/Unit/DataTransferObject/LoginHintDtoTest.php
luisfn: Can you please tell me the difference?

config/packages/test/lti1p3.yaml
```diff
@@ -0,0 +1,12 @@
+lti1p3:
+    key_chains:
+        platformKey:
+            key_set_name: "platformSet"
+            public_key: "file://%kernel.project_dir%/tests/Resources/secrets/public.key"
+            private_key: "file://%kernel.project_dir%/tests/Resources/secrets//private.key"
```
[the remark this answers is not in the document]
luisfn: done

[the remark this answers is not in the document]
luisfn: I'm afraid they are expected

docs/features.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### LTI 1.1 Configuration
+
+Configure the following environment variables according to the tool you are connecting to simple roster:
+
+- LTI_LAUNCH_PRESENTATION_RETURN_URL
+- LTI_LAUNCH_PRESENTATION_LOCALE
+- LTI_INSTANCE_LOAD_BALANCING_STRATEGY
+- LTI_OUTCOME_XML_NAMESPACE
+- LTI_VERSION (need to be set to `LTI-1p0`)
```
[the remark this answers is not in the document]
luisfn: Done

docs/features.md
```diff
@@ -0,0 +1,121 @@
...
+        - [LTI 1.3 Configuration](#lti-13-configuration)
+        - [LTI 1.3 Link Generation](#lti-13-link-generation-and-response)
+
+## LTI
+
+Learning Tools Interoperability (LTI) is an education technology specification developed by the IMS Global Learning Consortium. It specifies a method for a learning system to invoke and to communicate with external systems.
+
+Simple Roster is capable of generate links for the LTI versions `1.1` and `1.3`. In order configure which version you want to use, you need to set the `LTI_VERSION` environment variable. For a complete list of environment variable, please check please check [DevOps documentation - Lti Related Environment Variables](devops-documentation.md#lti-related-environment-variables).
+
+
```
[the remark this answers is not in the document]
luisfn: Doc was updated

docs/features.md
```diff
@@ -0,0 +1,121 @@
...
+- LTI_LAUNCH_PRESENTATION_RETURN_URL
+- LTI_LAUNCH_PRESENTATION_LOCALE
+- LTI_INSTANCE_LOAD_BALANCING_STRATEGY
+- LTI_OUTCOME_XML_NAMESPACE
+- LTI_VERSION (need to be set to `LTI-1p0`)
+
+create your LTI Instances using the [CLI](cli/ingester-command.md#examples)
+
+#### LTI 1.1 Link generation and Response
```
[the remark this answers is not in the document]
luisfn: I can do it, but then we won't have much here I think :)

docs/features.md
```diff
@@ -0,0 +1,121 @@
...
+    "context_title": "string",
+    "context_type": "string",
+    "roles": "string",
+    "user_id": 0,
+    "lis_person_name_full": "string",
+    "resource_link_id": 0,
+    "lis_outcome_service_url": "string",
+    "lis_result_sourcedid": 0,
+    "launch_presentation_return_url": "string",
+    "launch_presentation_locale": "it-IT"
```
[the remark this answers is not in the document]
luisfn: Applied

src/Lti/Exception/InvalidGroupException.php
```diff
@@ -0,0 +1,30 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ *  Copyright (c) 2020 (original work) Open Assessment Technologies S.A.
+ */
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Lti\Exception;
```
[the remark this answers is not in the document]
luisfn: Changed the logic to not include group on the login hint, so this exception will not be necessary

src/DataTransferObject/LoginHintDto.php
luisfn: Changed to have username and assignment id

src/Lti/Factory/Lti1p1RequestFactory.php
luisfn: The only change is that `checkIfAssignmentCanBeProcessed(Assignment $assignment)` remains on the service

src/Lti/Factory/Lti1p1RequestFactory.php
luisfn: It's not exactly mine, but removed :)

src/Lti/Factory/Lti1p3RequestFactory.php
luisfn: Renamed builder as well

src/Lti/Factory/Lti1p3RequestFactory.php
luisfn: don't see much gain on turn it on constructor parameter but fixed the cast usage. By the way, Why cast should not have spaces?

On https://www.php-fig.org/psr/psr-12/ it says `Type casting operators MUST NOT have any space within the parentheses` but does not mention after the parentheses.

src/Lti/Factory/Lti1p3RequestFactory.php
luisfn: Fixed

src/Lti/Factory/Lti1p3RequestFactory.php
luisfn: Added the LaunchPresentationClaim

src/Lti/Factory/LtiRequestFactory.php
luisfn: If I change it, I will need to add all the parameter from both factories on the constructor here. Are you sure I need to change it?

src/Lti/Factory/LtiRequestFactoryInterface.php
luisfn: Removed

src/Lti/Request/LtiRequest.php
luisfn: Done

src/Security/Lti/OidcUserAuthenticator.php
luisfn: removed

tests/Traits/KeyPairTestingTrait.php
luisfn: Done

tests/Traits/KeyPairTestingTrait.php
luisfn: Resolved since we could added a key pair, so no  trait anymore

tests/Unit/Lti/Extractor/LoginHintExtractorTest.php
luisfn: done

src/Lti/Factory/LtiRequestFactory.php
uncleempty: IMO - yes.
Let me explain a bit:
IMO - it would be better to have this as a factory class, which indeed is injected by all the args/services needed to build "versioned factories" (BTW, since they're not real factories, I would rename them as `Lti1p1RequestBuilder` / `Lti1p3RequestBuilder`). Even if it will increase the class complexity (which we will be trying to deal with for other services in next sprint), it would be more "cleaner" way as for me

.env
```diff
@@ -43,8 +43,14 @@ ASSIGNMENT_STATE_INTERVAL_THRESHOLD=P1D
 LTI_LAUNCH_PRESENTATION_LOCALE=en-EN
 LTI_LAUNCH_PRESENTATION_RETURN_URL=http://localhost:8000/index.html
 LTI_OUTCOME_XML_NAMESPACE=http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0
+# Possible values for LTI Version: 1.1.1 or 1.3.0
+LTI_VERSION=1.1.1
 # Possible values for LTI Instance load balancer strategy: username or userGroupId
 LTI_INSTANCE_LOAD_BALANCING_STRATEGY=username
+# Key used for LTI 1.3 link generation - Needed only when LTI_VERSION is set to 1.3.0
+LTI1P3_SERVICE_ENCRYPTION_KEY=lti1p3Key
```
ekkinox, line 51: you should leave this env surrounded this way:

```
###> oat-sa/bundle-lti1p3 ###
LTI1P3_SERVICE_ENCRYPTION_KEY=xxx
###< oat-sa/bundle-lti1p3 ###
```

this is how SF flex recipes detects that the .env file need update or not :)

ekkinox: also, dont hesitate to bring a bit more entropy in this key, it's goal remain to secure, predictable keys should be avoided, whatever environment we target

this applies to all similar cases

composer.json
```diff
@@ -16,6 +16,7 @@
         "league/csv": "^9.0",
         "nelmio/cors-bundle": "^2.0",
         "nesbot/carbon": "^2.11",
+        "oat-sa/bundle-lti1p3": "^3.0",
```
ekkinox, line 19: please req`"oat-sa/bundle-lti1p3": "^3.1",`, you'll need the scope compliant firewalls, released in `3.1.0`

see https://github.com/oat-sa/bundle-lti1p3/releases/tag/3.1.0 for details

config/packages/lti1p3.yaml
```diff
@@ -0,0 +1,31 @@
...
+            deep_linking_url: null
+    registrations:
+        demo:
+            client_id: "demo"
+            platform: "localPlatform"
+            tool: "demoTool"
+            deployment_ids:
+                - "1"
+            platform_key_chain: "platformKey"
+            tool_key_chain: null
```
ekkinox, line 29: you're the platform, when you register a tool, you need to give the`tool_key_chain` or `tool_jwks_url` (or both).
=> otherwise, your platform will never be able to validate tool originating tokens (401 forever)

config/packages/test/lti1p3.yaml
```diff
@@ -0,0 +1,8 @@
+lti1p3:
+    key_chains:
+        platformKey:
+            key_set_name: "platformSet"
+            public_key: "file://%kernel.project_dir%/config/secrets//test/public.key"
```
ekkinox, line 5: faulty path: `//`

config/routes/lti1p3.yaml
```diff
@@ -0,0 +1,11 @@
...
+    resource: '@Lti1p3Bundle/Resources/config/routing/jwks.yaml'
+
+lti1p3_message_platform:
+    resource: '@Lti1p3Bundle/Resources/config/routing/message/platform.yaml'
+
+lti1p3_message_tool:
+    resource: '@Lti1p3Bundle/Resources/config/routing/message/tool.yaml'
+
+lti1p3_service_platform:
+    resource: '@Lti1p3Bundle/Resources/config/routing/service/platform.yaml'
```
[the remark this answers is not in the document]
ekkinox: maybe I should bring them in the bundle docs, to ease other apps integrations.

once you documented them, can you ping me so I can PR bundle's doc ?

config/secrets/test/private.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
ekkinox, line 1: please NEVER push private keys (even test ones), I bet github bots should already have warned you :)

instead, in your readme, you can link to the doc about how to generate keys: https://github.com/oat-sa/bundle-lti1p3/blob/master/doc/quickstart/configuration.md#configure-a-keychain

config/services.yaml
```diff
@@ -15,6 +15,8 @@ parameters:
     app.cache.ttl.get_user_with_assignments: '%env(int:CACHE_TTL_GET_USER_WITH_ASSIGNMENTS)%'
     app.cache.ttl.lti_instance_list: '%env(int:CACHE_TTL_LTI_INSTANCES)%'
     app.lti.load_balancing_strategy: '%env(resolve:LTI_INSTANCE_LOAD_BALANCING_STRATEGY)%'
+    app.lti.version: '%env(resolve:LTI_VERSION)%'
+    app.lti.registration_id: '%env(resolve:LTI1P3_REGISTRATION_ID)%'
```
ekkinox, line 19: optional: what about `app.lti1p3.registration_id` since the concept of registrations applies to 1.3 ?

src/Lti/Exception/RegistrationNotFoundException.php
ekkinox: aren't the core `LtiExceptionInterface` enough ?

src/Lti/Factory/Lti1p3RequestFactory.php
```diff
@@ -0,0 +1,108 @@
...
+        if (!$registration) {
+            throw new RegistrationNotFoundException(sprintf('Registration %s not found.', $this->ltiRegistrationId));
+        }
+
+        $loginHint = new LoginHintDto(
+            (string)$assignment->getUser()->getUsername(),
+            (int)$assignment->getId(),
+        );
+
+        $message = $this->ltiRequestBuilder->buildLtiResourceLinkLaunchRequest(
```
ekkinox, line 81: seems a perfect usage of the builder :)

src/Lti/Factory/LtiRequestFactory.php
ekkinox: you can actually put the suggested builders (`Lti1p1RequestBuilder` and `Lti1p3RequestBuilder`) behind a `LtiRequestBuilderInterface`, and use `SF's service factories` to return the wanted instance based on your app configuration (will be autoresolved and autowired where needed).

imo, will ease maintenance, you can focus on business logic. 

see https://symfony.com/doc/current/service_container/factories.html#passing-arguments-to-the-factory-method

src/Security/Lti/OidcUserAuthenticator.php
ekkinox: actually you should let bubble up the exceptions, they will be catched and stamped 401 in the bundle automated endpoint. 

see https://github.com/oat-sa/bundle-lti1p3/blob/master/Action/Platform/Message/OidcAuthenticationAction.php#L48

good idea for the logger, always help

ekkinox (changes requested):
nice job !

few fixes to provide, but LTI 1.3 speaking, seems correct.

src/Lti/Extractor/LoginHintExtractor.php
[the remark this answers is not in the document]
luisfn: Not using group anymore

config/packages/lti1p3.yaml
ekkinox: as discussed, please just put `tool_jwks_url: "http://localhost:8888/lti1p3/.well-known/jwks/toolSet.json"`

tests/Unit/Lti/Factory/Lti1p3RequestFactoryTest.php
Skyld:
```
/**
     * Returns a matcher that matches when the method is executed exactly once.
     */
    public static function once(): InvokedCountMatcher
    {
        return new InvokedCountMatcher(1);
    }
```

config/services.yaml
Skyld: Fixed

src/Lti/Factory/LtiRequestFactory.php
Skyld: I agree, renamed class

src/Security/Lti/OidcUserAuthenticator.php
Skyld: Thx @ekkinox !

Skyld (approved):

src/Lti/Extractor/LoginHintExtractor.php
```diff
@@ -0,0 +1,56 @@
...
+     * @throws LogicException
+     */
+    public function extract(string $loginHint): LoginHintDto
+    {
+        $matches = [];
+
+        preg_match('/^(?P<username>.*)::(?P<assignmentId>.*)$/', $loginHint, $matches);
+
+        if (empty($matches)) {
+            throw new LogicException('Invalid Login hint format.');
```
ekkinox, line 40: imo, this is not `LogicException` (no code fix will solve the issue), but a `InvalidArgumentException` instead, wdyt ?

ekkinox (approved):

ekkinox:

uncleempty (approved):
I agree with @ekkinox comment about changing the exception thrown. But, in any case, it's a minor change, so approved from my side.
- [ ] New code is not subject to concurrency issues
- [ ] Feature is working correctly on my local machine

andreluizmachado (approved):
- [x] New code is not subject to concurrency issues (if applicable)

PR #173, opened 2020-11-27
[RO-79] Update Line Items via web hook

andreluizmachado:
This ticket is intended to implement the User Story: https://oat-sa.atlassian.net/browse/RO-79

### Before this implementation:
 - Line items cache is coupled with user's cache
 - There is no possibility to update line items unless you reingest it and warm up the cache again
### After this implementation:
 - Line items cache is decoupled with user's cache
 - Line items cache can be warmed up separately
 - A new endpoint was created to receive notifications about line item updates (**please check:** api_v1.yml)


src/EventListener/Doctrine/WarmUpLineItemListener.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->cacheProvider = $cacheProvider;
+        $this->lineItemCacheIdGenerator = $lineItemCacheIdGenerator;
+        $this->lineItemRepository = $lineItemRepository;
+    }
+
+    public function postUpdate(LineItem $lineItem): void
+    {
+        $id = (int)$lineItem->getId();
+        $this->cacheProvider->delete($this->lineItemCacheIdGenerator->generate($id));
+        $this->lineItemRepository->findById($id);
```
luisfn, line 65: Why do we need this `findById` in the end?

src/Repository/UserRepository.php
```diff
@@ -59,7 +60,10 @@ public function __construct(
     }
 
     /**
-     * @throws InvalidUsernameException
+     * @param string $username
+     *
+     * @return User
+     *
```
luisfn, line 66: I think we can have only the `throws` section

src/Repository/UserRepository.php
```diff
@@ -88,7 +92,11 @@ public function findByUsernameWithAssignments(string $username): User
     }
 
     /**
-     * @throws InvalidArgumentException
+     * @param int $limit
+     * @param int|null $lastUserId
+     * @param FindUserCriteria|null $criteria
+     * @return UsernameResultSet
```
luisfn, line 98: Same as above

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
```diff
@@ -0,0 +1,70 @@
...
+    public function apply(Request $request, ParamConverter $configuration): bool
+    {
+        $this->updateLineItemValidator->validate($request);
+
+        $responseBody = json_decode($request->getContent(), true);
+        $eventsRaw = $responseBody['events'];
+        $events = [];
+
+        foreach ($eventsRaw as $event) {
+            $events[] = new UpdateLineItemDto(
```
luisfn, line 52: Maybe rename `UpdateLineItemDto` to `UpdateLineItemEvent`?
Seems more reasonable to me since it mostly contain event data. What do you think?

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        $request->attributes->set($configuration->getName(), new UpdateLineItemCollection(...$events));
+
+        return true;
+    }
+
+    public function supports(ParamConverter $configuration): bool
+    {
+        return UpdateLineItemCollection::class === $configuration->getClass();
```
luisfn, line 68: If you agree to rename the DTO, this collection can be renamed also

src/Request/Validator/UpdateLineItemValidator.php
```diff
@@ -0,0 +1,130 @@
...
+            );
+        }
+
+        $errors = $this->validator->validate($responseBody, $this->getConstraints());
+
+        if ($errors->count() < 1) {
+            return;
+        }
+
+        $errorsRaw = [];
```
luisfn, line 62: Maybe rename to `$rawErrors` (just being picky here)

src/Request/Validator/UpdateLineItemValidator.php
```diff
@@ -0,0 +1,130 @@
...
+            );
+        }
+
+        throw new BadRequestHttpException(
+            sprintf(
+                'Invalid Request Body: %s',
+                implode(
+                    " ",
+                    $errorsRaw
+                )
```
luisfn, line 78: We don't need multiple lines for the implode section. Also maybe separate it with comma

src/Request/Validator/UpdateLineItemValidator.php
```diff
@@ -0,0 +1,130 @@
...
+                    'Invalid JSON request body received. Error: %s.',
+                    $jsonException->getMessage()
+                ),
+                $jsonException
+            );
+        }
+
+        $errors = $this->validator->validate($responseBody, $this->getConstraints());
+
+        if ($errors->count() < 1) {
```
luisfn, line 58: Just preference but would use `if ($errors->count() === 0) {`

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,96 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Service\WebHook;
+
+use _HumbugBoxf99c1794c57d\Symfony\Component\Console\Exception\LogicException;
```
luisfn, line 25: I think this is not the exception you are looking for :)

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        $lineItems = $this->lineItemRepository->findBy(
+            [
+                'slug' => (array)$slugs
+            ]
+        );
+
+        foreach ($lineItems as $lineItem) {
+            $duplicatedEvents = $knownUpdates
```
luisfn, line 74: IMO this confirms the `*Dto`should be renamed to `*Event`

Something like: `$duplicatedEvents = $knownUpdateEvents->filter()`

tests/Fixtures/3LineItems.yml
```diff
@@ -0,0 +1,12 @@
+OAT\SimpleRoster\Entity\LineItem:
+    line_item_{1..2}]:
+            label: 'line_item_<current()>'
+            uri: 'https://test.taocloud.fr/__n/<current()>/'
+            slug: 'slug-<current()>'
+    line_item_qqy:
```
luisfn, line 6: Does `qqy` have any meaning or just random?

luisfn (changes requested):
Good work with the PR. I'm not the best when it comes to the warmup topic but could get the idea. Have just some small remarks.

As a last suggestion, would rename the PR to `[Task Id] Title` format. Maybe we can sync it between team members.

tests/Fixtures/3LineItems.yml
andreluizmachado: yes, It is :)

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: I do not think that it's a good idea to bring these events to our bounded context. In our context, we have line items, not these events. if we rename it, it can cause confusion inside the project with the proper events that we can have into our application.

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: Thanks @luisfn :)

src/EventListener/Doctrine/WarmUpLineItemListener.php
andreluizmachado: This is warming up the cache again by using new database data.

src/Command/Cache/LineItemCacheWarmerCommand.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @return int
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        try {
+            $this->symfonyStyle->comment('Executing cache warmup...');
```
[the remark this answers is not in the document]
andreluizmachado: To keep consistency with the other commands, I'm going to keep it.

src/Command/Cache/LineItemCacheWarmerCommand.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->resultCacheImplementation = $resultCacheImplementation;
+        $this->lineItemRepository = $lineItemRepository;
+        $this->lineItemCacheIdGenerator = $lineItemCacheIdGenerator;
+    }
+
+    protected function configure(): void
+    {
+        parent::configure();
+
+//        $this->addBlackfireProfilingOption();
```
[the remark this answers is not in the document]
andreluizmachado: Yes, when @Skyld PR is merged.

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
andreluizmachado: same as described here: https://github.com/oat-sa/simple-roster/pull/173#discussion_r533292698

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
andreluizmachado: same as described here: https://github.com/oat-sa/simple-roster/pull/173#discussion_r533292698

src/Request/Validator/UpdateLineItemValidator.php
```diff
@@ -0,0 +1,122 @@
...
+                $error->getMessage()
+            );
+        }
+
+        throw new BadRequestHttpException(sprintf('Invalid Request Body: %s', implode(" ", $rawErrors)));
+    }
+
+    /**
+     * @return Assert\Collection
+     */
```
Skyld, line 77: You can remove this docbloc

src/Request/Validator/UpdateLineItemValidator.php
```diff
@@ -0,0 +1,122 @@
...
+namespace OAT\SimpleRoster\Request\Validator;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+use Throwable;
+
+class UpdateLineItemValidator
```
Skyld, line 32: Please add unit tests

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
```diff
@@ -0,0 +1,70 @@
...
+
+use DateTimeImmutable;
+use OAT\SimpleRoster\Request\Validator\UpdateLineItemValidator;
+use OAT\SimpleRoster\WebHook\UpdateLineItemCollection;
+use OAT\SimpleRoster\WebHook\UpdateLineItemDto;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
+use Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class UpdateLineItemWebHookParamConverter implements ParamConverterInterface
```
Skyld, line 33: Please add unit tests

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Service\WebHook;
+
+use Doctrine\ORM\EntityManagerInterface;
+use OAT\SimpleRoster\Repository\LineItemRepository;
+use OAT\SimpleRoster\WebHook\UpdateLineItemCollection;
+use OAT\SimpleRoster\WebHook\UpdateLineItemDto;
+
+class UpdateLineItemsService
```
Skyld, line 30: Please add unit tests

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    /** @var LineItemRepository */
+    private $lineItemRepository;
+
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    public function __construct(LineItemRepository $lineItemRepository, EntityManagerInterface $entityManager)
```
Skyld, line 38: Please add logging

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+
+    public function handleUpdates(UpdateLineItemCollection $collection): UpdateLineItemCollection
+    {
+        $knownUpdates = $collection
+            ->filter(
+                function (UpdateLineItemDto $dto): bool {
+                    return $dto->getName() === UpdateLineItemDto::NAME;
+                }
+            )
+            ->setStatus(UpdateLineItemDto::STATUS_ERROR);
```
Skyld, line 52: If we ignore all the events we are not interested in (see my comment above) I think we can remove this.

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+        $slugs = $knownUpdates
+            ->map(
+                function (UpdateLineItemDto $dto): string {
+                    return (string)$dto->getSlug();
+                }
+            );
+
+        if (count((array)$slugs) < 1) {
+            return $collection;
+        }
```
Skyld, line 63: Here it's visible what is one of my concerns regarding this "collection pipeline" approach. I actually like the idea 👍 , but as you can see here with mapping the collection you just suddenly lost your types. You would expect `string[]`, but you get an iterator instead, forcing you to explicitly case it into an array. If you would encapsulate the mapping logic inside the collection, you could work with less code here and with proper types. What do you think?

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+
+        $lineItems = $this->lineItemRepository->findBy(
+            [
+                'slug' => (array)$slugs
+            ]
+        );
+
+        foreach ($lineItems as $lineItem) {
+            $duplicatedEvents = $knownUpdates
+                ->filter(
```
Skyld, line 73: There are no tests to check it but I believe you are not checking an edge case when the filtering returns an empty array.

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+                    function (UpdateLineItemDto $dto) use ($lineItem): bool {
+                        return $dto->getSlug() === $lineItem->getSlug();
+                    }
+                );
+
+            $dto = $duplicatedEvents->findLastByTriggeredTime();
+
+            if (null !== $dto) {
```
Skyld, line 81: At this point I think we should not have nullable checks in the code.

- We should know way before if the collection is empty. If empty, why bothering with all these logic above?
- Why does the collection contain unrelevant events? Let's ignore them in the param converter.

src/Service/WebHook/UpdateLineItemsService.php
```diff
@@ -0,0 +1,94 @@
...
+                $lineItem->setUri($dto->getLineItemUri());
+                $this->entityManager->persist($lineItem);
```
Skyld, line 83: Why are we using the entity manager to persist? We have already the repository for it, we can `save()` the line items.

src/WebHook/UpdateLineItemCollection.php
```diff
@@ -0,0 +1,96 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\WebHook;
+
+use ArrayIterator;
+use Closure;
+use Iterator;
+use IteratorAggregate;
+
+class UpdateLineItemCollection implements IteratorAggregate
```
Skyld, line 30: Please add unit tests

src/WebHook/UpdateLineItemDto.php
```diff
@@ -0,0 +1,104 @@
...
+use DateTimeInterface;
+use JsonSerializable;
+
+class UpdateLineItemDto implements JsonSerializable
+{
+    public const STATUS_ACCEPTED = 'accepted';
+    public const STATUS_ERROR = 'error';
+    public const STATUS_IGNORED = 'ignored';
+
+    public const NAME = 'RemoteDeliveryPublicationFinished';
```
Skyld, line 34: This is TAO specific data, it belongs to a webhook specific class.

src/WebHook/UpdateLineItemDto.php
```diff
@@ -0,0 +1,104 @@
...
+namespace OAT\SimpleRoster\WebHook;
+
+use DateTimeInterface;
+use JsonSerializable;
+
+class UpdateLineItemDto implements JsonSerializable
+{
+    public const STATUS_ACCEPTED = 'accepted';
+    public const STATUS_ERROR = 'error';
+    public const STATUS_IGNORED = 'ignored';
```
Skyld, line 32: This is TAO specific data, it belongs to a webhook specific class.

src/WebHook/UpdateLineItemDto.php
```diff
@@ -0,0 +1,104 @@
...
+    public const NAME = 'RemoteDeliveryPublicationFinished';
+
+    /** @var string */
+    private $id;
+
+    /** @var string */
+    private $name;
+
+    /** @var string|null */
+    private $slug;
```
Skyld, line 43: Why is it nullable? It should not be.

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+        $this->setUpDatabase();
+        $this->loadFixtureByFilename('userWithReadyAssignment.yml');
+
+        $this->setUpTestLogHandler();
+    }
+
+    /**
+     * @dataProvider provideWrongRequestBodies
+     * @param array|null $requestBody
+     * @param string $expectedMessage
```
Skyld, line 73: You can remove these lines

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+                            'triggeredTimestamp' => 1565602371,
+                            'eventData' => [
+                                'alias' => 'qti-interactions-delivery',
+                            ],
+                        ],
+                    ],
+                ],
+                'expectedMessage' => $missingDeliveryUri,
+            ],
+            'aliasWrongType' => [
```
Skyld, line 284: WrongAliasType

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+        $this->setUpTestLogHandler();
+    }
+
+    /**
+     * @dataProvider provideWrongRequestBodies
+     * @param array|null $requestBody
+     * @param string $expectedMessage
+     * @throws JsonException
+     */
+    public function testItThrowsUnauthorizedHttpExceptionIfRequestApiKeyIsInvalid(
```
Skyld, line 76: If request api key is invalid? Copy paste? :)

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+                    ],
+                ],
+                [
+                    'eventId' => '52a3de8dd0f270fd193f9f4bff05232c',
+                    'eventName' => 'RemoteDeliveryPublicationFinished',
+                    'triggeredTimestamp' => 1565602390,
+                    'eventData' => [
+                        'alias' => 'lineItemSlug',
+                        'deliveryURI' => 'https://docker.localhost/ontologies/tao.rdf#RightOne',
+                    ],
```
Skyld, line 342: 👍  nice, but you could play also with the order of events to see if you properly filter by last triggered at date

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+                    'eventId' => '52a3de8dd0f270fd193f9f4bff05232c',
+                    'status' => 'accepted'
+                ]
+            ],
+            $decodedResponse
+        );
+    }
+
+
+    public function testItAcceptsRequestButIgnoreUnknownsEvents(): void
```
Skyld, line 174: .....UnknownEvents

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
```diff
@@ -0,0 +1,365 @@
...
+                [
+                    'eventId' => '52a3de8dd0f270fd193f9f4bff05232f',
+                    'status' => 'ignored'
+                ]
+            ],
+            $decodedResponse
+        );
+    }
+
+    public function provideWrongRequestBodies(): array
```
Skyld, line 205: Well done 👍

tests/Functional/Command/Cache/LineItemCacheWarmerCommandTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $commandTester = new CommandTester($application->find(LineItemCacheWarmerCommand::NAME));
+
+        self::assertSame(1, $commandTester->execute([], ['capture_stderr_separately' => true]));
+        self::assertStringContainsString(
+            '[ERROR] An unexpected error occurred: Yaaay',
+            $commandTester->getDisplay()
+        );
+    }
+
+    public function testItDisplaysWarningIfThereAreNoLtiInstancesIngested(): void
```
Skyld, line 101: Copy paste power :) line item, not lti instance

Skyld (changes requested):

config/doctrine/LineItem.orm.xml
```diff
@@ -35,5 +35,10 @@
                 <option name="default">0</option>
             </options>
         </field>
+        <entity-listeners>
+            <entity-listener class="OAT\SimpleRoster\EventListener\Doctrine\WarmUpLineItemListener">
```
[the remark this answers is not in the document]
andreluizmachado: Good Idea. Thanks @Skyld

src/Action/WebHook/UpdateLineItemsAction.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->responder = $responder;
+        $this->service = $service;
+    }
+
+    public function __invoke(UpdateLineItemCollection $collection): Response
+    {
+        return $this->responder->createJsonResponse(
+            $this->service->handleUpdates($collection),
+            Response::HTTP_OK
```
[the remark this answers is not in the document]
andreluizmachado: Unfortunately not, we can only use 2xx for this case otherwise it'll result in a request retries and the best one that we have on 2xx family is the 200.

src/Repository/LineItemRepository.php
```diff
@@ -48,4 +60,16 @@ public function findAllAsCollection(): LineItemCollection
 
         return $collection;
     }
+
+    public function findById(int $id): ?LineItem
```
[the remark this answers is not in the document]
andreluizmachado: I'm consistent. IMO find methods should NOT return this exception unless you specify it on the method name. This shouldn't be an exception because it's expected to do not have such information on the database. If you take a look at the `find` or `findOneBy` Methods, they do not return exceptions.

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
```diff
@@ -0,0 +1,70 @@
...
+        $this->updateLineItemValidator->validate($request);
+
+        $responseBody = json_decode($request->getContent(), true);
+        $eventsRaw = $responseBody['events'];
+        $events = [];
+
+        foreach ($eventsRaw as $event) {
+            $events[] = new UpdateLineItemDto(
+                (string)$event['eventId'],
+                (string)$event['eventName'],
```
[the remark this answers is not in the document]
andreluizmachado: About the logging, I do not know what you mean I checked other ParamConverters on this application and those aren't logging anything.
About the collection. It's not about the filter, it's about representing in a OOP a collection of objects. This class doesn't have the responsibility to filter any data, only to converter it. If we're filtering something here it's a violation of single responsibility.
About The DTO names, there is one important thing here, this is a webhook which it doesn't belong to our context and I wouldn't even consider it as REST endpoint concept. As you can see, this request was not designed thinking about our context (which'd be a PUT/PATCH of line items). It was designed to be more generic and let the client to decide what to do with the information. That's why I decided to put it on the DTO name because that's the purpose of a DTO object. I'm just encapsulating data to be represented between two systems.

src/Request/Validator/UpdateLineItemValidator.php
andreluizmachado: Why?

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
andreluizmachado: Why?

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: Could you explain to me the criteria? because a see that most of the classes are converted only by functional/integration tests. EX: UserCacheWarmerService, Ingester commands, ...

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: I do not like the idea of doing it on ParamConverter. It's a single Responsibility violation.

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: The types were not lost, it's just a matter of taste. I can change it to an array (string[]) if you want. It was me who add an iterator instead on the UpdateLineItemCollection::map. It was more to use OOP instead of primitive data type.

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: There is no way to this be empty here.

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: Again Single Responsible Violation and I just added it because of phpstan there is no change to this happen here.

src/Service/WebHook/UpdateLineItemsService.php
andreluizmachado: well noted. Thanks @Skyld.

src/WebHook/UpdateLineItemDto.php
andreluizmachado: The webhook doesn`t have the obligation to send it. this data is not mandatory data on the Authoring side.

src/WebHook/UpdateLineItemDto.php
andreluizmachado: But this is the webhook specific class that we have unless we use the UpdateLineItemService, there is no other one.

src/WebHook/UpdateLineItemCollection.php
andreluizmachado: https://github.com/oat-sa/simple-roster/pull/173#discussion_r534155509

tests/Functional/Action/WebHook/UpdateLineItemsActionTest.php
andreluizmachado: Thank you : 😊

tests/Functional/Command/Cache/LineItemCacheWarmerCommandTest.php
andreluizmachado: Thank you :)

tests/Unit/EventListener/Doctrine/WarmUpLineItemListenerTest.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Doctrine\ORM\Configuration;
+use Doctrine\ORM\EntityManagerInterface;
+use OAT\SimpleRoster\EventListener\Doctrine\WarmUpLineItemCacheListener;
+use OAT\SimpleRoster\Exception\DoctrineResultCacheImplementationNotFoundException;
+use OAT\SimpleRoster\Generator\LineItemCacheIdGenerator;
+use OAT\SimpleRoster\Repository\LineItemRepository;
+use PHPUnit\Framework\TestCase;
+
+class WarmUpLineItemListenerTest extends TestCase
```
luisfn, line 33: Please rename the test also

src/Command/Cache/LineItemCacheWarmerCommand.php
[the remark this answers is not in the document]
luisfn: So I would change on the specific "warmers" as well, but it's more personal choice. This is the output to an user, so I think the more information the better.

tests/Unit/EventListener/Doctrine/WarmUpLineItemListenerTest.php
andreluizmachado: Thanks. Fixed. :)

src/Generator/LineItemCacheIdGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Generator;
+
+class LineItemCacheIdGenerator
```
luisfn, line 25: Maybe this could be a method from the entity, like `getCacheId()`. My only concern is that we would need as many generators as the entities we have.

andreluizmachado: To keep consistency, I did exactly like UserCacheIdGenerator class. :)

luisfn: It's ok, just don't see the gain on having these generators. It's one more dependency.

tests/Unit/EventListener/Doctrine/WarmUpLineItemCacheListenerTest.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Doctrine\ORM\Configuration;
+use Doctrine\ORM\EntityManagerInterface;
+use OAT\SimpleRoster\EventListener\Doctrine\WarmUpLineItemCacheListener;
+use OAT\SimpleRoster\Exception\DoctrineResultCacheImplementationNotFoundException;
+use OAT\SimpleRoster\Generator\LineItemCacheIdGenerator;
+use OAT\SimpleRoster\Repository\LineItemRepository;
+use PHPUnit\Framework\TestCase;
+
+class WarmUpLineItemCacheListenerTest extends TestCase
```
Skyld, line 33: Let's add these test cases:

- Check if it is a doctrine entity listener
- Test postUpdate method if it does what is expected

tests/Unit/Request/ParamConverter/UpdateLineItemWebHookParamConverterTest.php
```diff
@@ -0,0 +1,130 @@
...
+use PHPUnit\Framework\TestCase;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+class UpdateLineItemWebHookParamConverterTest extends TestCase
+{
+    /** @var UpdateLineItemWebHookParamConverter */
+    private $sut;
```
Skyld, line 40: I know what it means but let's try to avoid using abbreviations whenever it's possible. Let's use `subject`

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
```diff
@@ -0,0 +1,250 @@
...
+        $this->sut = new UpdateLineItemsService(
+            $this->lineItemRepository,
+            $this->entityManager,
+            $this->logger
+        );
+    }
+
+    public function testItIgnoresUnknownUpdates(): void
+    {
+        $this->lineItemRepository->expects($this->never())->method('findBy');
```
Skyld, line 73: self::never (please fix in other methods as well, won't comment on all of them)

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
```diff
@@ -0,0 +1,250 @@
...
+                'qti-interactions-delivery'
+            )
+        );
+
+        $result = $this->sut->handleUpdates($updateLineItemCollection);
+
+        /** @var UpdateLineItemDto $updateLineItemDto */
+        $updateLineItemDto = $result->getIterator()[0];
+
+        $this->assertSame('ignored', $updateLineItemDto->getStatus());
```
Skyld, line 91: self::assertSame

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
```diff
@@ -0,0 +1,250 @@
...
+                [
+                    'slug' => ['qti-interactions-delivery']
+                ]
+            )
+            ->willReturn(
+                [
+                    (new LineItem())
+                        ->setSlug('qti-interactions-delivery')
+                ]
+            );
```
Skyld, line 108:
```
$this->lineItemRepository
        ->expects(self::once())
        ->method('findBy')
        ->with(['slug' => ['qti-interactions-delivery']])
        ->willReturn([(new LineItem())->setSlug('qti-interactions-delivery')]);
```

5 lines instead of 13

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
```diff
@@ -0,0 +1,250 @@
...
+            ->with(
+                [
+                    'slug' => ['qti-interactions-delivery','qti-interactions-delivery']
+                ]
+            )
+            ->willReturn(
+                [
+                    $lineItem
+                ]
+            );
```
Skyld, line 157: Same issue here (breaking everything into new lines)

tests/Unit/Webhook/UpdateLineItemCollectionTest.php
```diff
@@ -0,0 +1,156 @@
...
+use DateTimeImmutable;
+use Exception;
+use Iterator;
+use OAT\SimpleRoster\WebHook\UpdateLineItemCollection;
+use OAT\SimpleRoster\WebHook\UpdateLineItemDto;
+use PHPUnit\Framework\TestCase;
+
+class UpdateLineItemCollectionTest extends TestCase
+{
+    public function testItThrowsExceptionWhenEmptyCollection(): void
```
Skyld, line 34: Test case name doesn't describe what is tested: testItThrowsExceptionWhenEmptyCollectionIsFilteredByTriggeredTime(

tests/Unit/Webhook/UpdateLineItemCollectionTest.php
```diff
@@ -0,0 +1,156 @@
...
+                'http://i.o',
+                (new DateTimeImmutable())->setTimestamp(1565602371)
+            ),
+            new UpdateLineItemDto(
+                '22',
+                'test',
+                'http://i.o',
+                (new DateTimeImmutable())->setTimestamp(1565602380)
+            )
+        );
```
Skyld, line 58: Why not switching the order so you can actually see if it's "finding" it?

tests/Unit/Webhook/UpdateLineItemCollectionTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+        $dto = $collection->findLastByTriggeredTimeOrFail();
+        $this->assertEquals('22', $dto->getId());
```
Skyld, line 61: Please use them as static (all occurrences)

Skyld (changes requested):

config/doctrine/Assignment.orm.xml
```diff
@@ -32,6 +32,7 @@
                 <option name="default">0</option>
             </options>
         </field>
+        <field name="lineItemId" type="integer" column="line_item_id" precision="0" scale="0" nullable="false"/>
```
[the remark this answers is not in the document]
andreluizmachado: I'm not sure if we can remove it. Let me check.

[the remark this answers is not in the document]
andreluizmachado: I can't: Return value of OAT\\SimpleRoster\\Entity\\Assignment::getLineItemId() must be of the type int, null returned

docs/features/update-line-items-webhook.md
```diff
@@ -0,0 +1,182 @@
...
+### Endpoint Descriptions
+
+You can find the endpoint description by accessing the open [api specification](../../openapi/api_v1.yml)
+
+### WebHook Payload Schema Definition
+
+These are the request and schema definitions for the related webhook:
+
+#### Request
+```
```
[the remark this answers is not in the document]
andreluizmachado: What's the reason for this case?

docs/features/update-line-items-webhook.md
```diff
@@ -0,0 +1,182 @@
...
+		}
+	]
+}'
+```
+
+#### Attribute Descriptions
+
+| Attribute                 | Description                                                                                                                                                                                                                 |
+| --------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| eventId                   | The event identifier. The format is described in the diagram. It is used by the simple list only to return it in the webhook response.                                                                                      |
```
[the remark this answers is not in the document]
andreluizmachado: Thanks @Skyld

src/Command/Cache/LineItemCacheWarmerCommand.php
```diff
@@ -0,0 +1,151 @@
...
+                        'cacheTtl' => number_format($this->lineItemCacheTtl),
+                    ]
+                );
+            }
+
+            if ($lineItemCollection->isEmpty()) {
+                $this->symfonyStyle->warning('There are no Line Items found in the database.');
+
+                return 0;
+            }
```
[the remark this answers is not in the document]
andreluizmachado: Yes. Thanks :)

src/Command/Cache/LineItemCacheWarmerCommand.php
```diff
@@ -0,0 +1,151 @@
...
+            /** @var LineItem $lineItem */
+            foreach ($lineItemCollection as $lineItem) {
+                $id = (int)$lineItem->getId();
+                $lineItemCacheKey = $this->lineItemCacheIdGenerator->generate($id);
+
+                $this->resultCacheImplementation->delete($lineItemCacheKey);
+                $this->lineItemRepository->findById($id);
+
+                $this->logger->info(
+                    sprintf('Result cache for Line Item Id %d have been successfully warmed up.', $id),
```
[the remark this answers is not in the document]
andreluizmachado: maybe the opposite no? Just because double quotes have less performance. So I'd propose to always use single quotes instead. What do you think?

src/EventListener/Doctrine/LineItemLoaderListener.php
```diff
@@ -0,0 +1,49 @@
...
+    public function postLoad(Assignment $assignment): void
+    {
+        $lineItem = $this->lineItemRepository->findById($assignment->getLineItemId());
+
+        //we're forcing the doctrine to load the line item by using cache instead of database query
+        if (null !== $lineItem) {
+            $assignment->setLineItem(
+                $lineItem
+            );
+        }
```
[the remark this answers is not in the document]
andreluizmachado: I believe that we have to use consistent return points when the situation applies it. In this case, it doesn't apply it's expected that a find method can not find the resource and this is not an exception. If you'd throw an exception here you're basically saying that a normal situation is now an exception.

tests/Unit/Request/Validator/UpdateLineItemValidatorTest.php
```diff
@@ -0,0 +1,118 @@
...
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+class UpdateLineItemValidatorTest extends TestCase
+{
+    /** @var MockObject|ValidatorInterface */
+    private $validator;
+
+    /** @var UpdateLineItemValidator */
+    private $sut;
```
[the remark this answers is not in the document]
andreluizmachado: ok. It'd be very kind if you could explain why?

tests/Unit/Request/Validator/UpdateLineItemValidatorTest.php
```diff
@@ -0,0 +1,118 @@
...
+        );
+        $this->sut = new UpdateLineItemValidator(
+            $this->validator
+        );
+    }
+
+    public function testItValidatesSuccessfull(): void
+    {
+        $request = $this->createMock(Request::class);
+        $request->expects($this->once())
```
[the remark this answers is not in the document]
andreluizmachado: Why?

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
andreluizmachado: What's the gain by doing this? Lines of code? For code, readability is better like it's. there is no sense in what you are saying. if you want to avoid lines, then you contradicting yourself by putting the expects(self::once()) into another line instead of keeping it the same.

tests/Unit/Webhook/Service/UpdateLineItemsServiceTest.php
andreluizmachado: There is no issue here. please check my comment above. https://github.com/oat-sa/simple-roster/pull/173#discussion_r537317534

src/WebHook/Service/UpdateLineItemsService.php
```diff
@@ -0,0 +1,135 @@
...
+                }
+            );
+
+        if (count($slugs) < 1) {
+            return $collection;
+        }
+
+        $lineItems = $this->lineItemRepository->findBy(
+            [
+                'slug' => (array)$slugs
```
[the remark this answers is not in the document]
andreluizmachado: Thank You 🙏

src/Request/ParamConverter/UpdateLineItemWebHookParamConverter.php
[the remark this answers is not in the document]
andreluizmachado: - If you want to log the request payload, then I'd go with a generic solution for all the endpoints. All of them do not log the payload. I do not see the point of only log this one. So my suggestion would be to have something like:
https://github.com/uniplaces/request-logger-bundle
- Collection: It's about avoiding primitive types by representing data and programming logic by using Object-oriented Program. There lots of benefits to using collections. Examples: logic encapsulated (you can't do it inside of primitive types), Code Responsibility (all of those logic operations are related to a specific object type, in this case, the DTOs), ....
- DTO: No this class name it's not about line item, I'm not mixing the domains here, this is exactly what I'm avoiding. I do not want to expose these events inside of our context, but I need to have these events data coming from tao

src/Repository/LineItemRepository.php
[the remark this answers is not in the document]
Skyld: Let's fix it based on our discussion yesterday

src/WebHook/Service/UpdateLineItemsService.php
```diff
@@ -0,0 +1,135 @@
...
+                        'There are duplicated updates on the request. All of them will be ignore except update id %s. ',
+                        $dto->getId()
+                    )
+                );
+            }
+
+            $lineItem->setUri($dto->getLineItemUri());
+
+            $this->logger->info(
+                sprintf('The line item id %d was updated', $lineItem->getId()),
```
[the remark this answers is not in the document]
Skyld: Yeah sorry, %d can be used ofc.

src/WebHook/Service/UpdateLineItemsService.php
```diff
@@ -0,0 +1,135 @@
...
+            $dto->setStatus(UpdateLineItemDto::STATUS_ACCEPTED);
+        }
+
+        /** @var UpdateLineItemDto $knownUpdate */
+        foreach ($knownUpdates as $knownUpdate) {
+            if ($knownUpdate->getStatus() === UpdateLineItemDto::STATUS_ERROR) {
+                $this->logger->error(
+                    sprintf('Impossible to update the line item. The slug %s does not exist.', $knownUpdate->getSlug()),
+                    [
+                        'updateId' => $knownUpdate->getId()
```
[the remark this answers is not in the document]
Skyld: I don't understand what do you mean, let's discuss this

Skyld (changes requested):

src/Generator/LineItemCacheIdGenerator.php
Skyld: Separation of concerns, Luis.

config/doctrine/Assignment.orm.xml
```diff
@@ -40,13 +41,18 @@
                 <join-column name="user_id" referenced-column-name="id"/>
```
Skyld, line 41: I know that it's not related to your work but could you please add `nullable="false"` here? Thanks

Skyld (changes requested):

Skyld (approved):

luisfn (approved):
**Review Checklist**
- [ ] New code is not subject to concurrency issues (if applicable)

PR #176, opened 2020-12-03
[RO-46] Replace PHPSESSID with JWT

uncleempty:
Ticket: https://oat-sa.atlassian.net/browse/RO-46

### Changed:
- Auth flow for api endpoints - now SR is using JWT tokens to authorize requests to API endpoints

### Added:
- Refresh JWT token endpoint


src/Action/Security/RefreshJWTTokenAction.php
```diff
@@ -0,0 +1,98 @@
...
+        if (!isset($decodedRequestBody['refreshToken'])) {
+            throw new BadRequestHttpException("Missing 'refreshToken' in request body.");
+        }
+
+        try {
+            $refreshToken = (new Parser())->parse($decodedRequestBody['refreshToken']);
+            $this->tokenVerifier->isValid($refreshToken);
+
+            $username = $refreshToken->getClaim('username');
+            $user = $this->userRepository->findOneByUsername($username);
```
andreluizmachado, line 80: You can not assume all of the exceptions here are invalid tokens. We can also have internal server errors here.

src/Action/Security/RefreshJWTTokenAction.php
```diff
@@ -0,0 +1,98 @@
...
+        int $accessTokenTtl
+    ) {
+        $this->manager = $manager;
+        $this->userRepository = $userRepository;
+        $this->tokenVerifier = $tokenVerifier;
+        $this->responder = $responder;
+        $this->accessTokenTtl = $accessTokenTtl;
+    }
+
+    public function __invoke(Request $request): Response
```
andreluizmachado, line 68: I'd encourage you to use the param converter feature from the Symfony application and pass the token directly on the invoke. Example:
 ```
public function __invoke(Lcobucci\JWT\Token $token): Response {
....
}
```
```
RefreshTokenParamConverter.php
class RefreshTokenParamConverter
{
...
    public function apply(Request $request, ParamConverter $configuration): bool
    {
         //validations here
         
         $request->attributes->set($configuration->getName(), $token);
        ....
    }
}

src/Action/Security/RefreshJWTTokenAction.php
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        try {
+            $refreshToken = (new Parser())->parse($decodedRequestBody['refreshToken']);
+            $this->tokenVerifier->isValid($refreshToken);
+
+            $username = $refreshToken->getClaim('username');
+            $user = $this->userRepository->findOneByUsername($username);
+        } catch (\Throwable $exception) {
+            throw new ConflictHttpException('Invalid token.');
```
andreluizmachado, line 82: Can we improve this message by adding more details?

src/Entity/User.php
```diff
@@ -117,6 +117,23 @@ public function getAssignments(): Collection
...
 
+    /**
+     * @return Collection|Assignment[]
+     */
+    public function getCancellableAssignments(): Collection
+    {
+        $assignments = $this->getAssignments();
+
+        $list = new ArrayCollection();
+        foreach ($assignments as $assignment) {
```
andreluizmachado, line 128:
```
$assignments->filter(function(Assignment $assignment): bool { return $assignment->isCancellable()})
```

andreluizmachado: Another thing, why don't you bring only the assignments that you need instead of brings everything and filter after?

andreluizmachado: Another thing, did you check if this is getting the data from the cache layer instead of database?

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+    public function testItThrowsExceptionOnInvalidToken(): void
+    {
+        $subject = new JwtTokenAuthenticator(
+            $this->createMock(AuthorizationHeaderTokenExtractor::class),
+            $this->createMock(JwtTokenVerifier::class),
+            $this->createMock(SerializerResponder::class)
+        );
+
+        $tokenMock = $this->getMockBuilder(Token::class)
```
andreluizmachado, line 98: Suggestion: use $this->createMock instead

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+            ->getMock();
+
+        $tokenMock->expects($this->any())->method('verify')->willReturn(false);
```
andreluizmachado, line 101: should be self::once, no?

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $userProviderMock = $this->getMockBuilder(UserProviderInterface::class)
+            ->getMock();
+
+        $this->expectException(AuthenticationException::class);
```
andreluizmachado, line 106: I'd move it to the beginning of the method

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->expectExceptionMessage('Invalid token.');
+
+        // @codingStandardsIgnoreStart
```
andreluizmachado, line 109: can we avoid it?

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
andreluizmachado: Because I do not believe that you need the original constructor, clone methods in this case.

tests/Unit/Security/Authenticator/JwtTokenAuthenticatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $subject->getUser('eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwidXNlcm5hbWUiOiJ1c2VyMSIsImlhdCI6MTUxNjIzOTAyMn0.Fc59qPPNFvtlvwXdTyWTe8Uz6uu-EnQncjGow1RooHM', $userProviderMock);
```
andreluizmachado, line 110: can u please indent the code?

andreluizmachado (changes requested):
Hi, nice job by the way.
I have some comments, please check it.

config/services.yaml
```diff
@@ -38,6 +43,11 @@ services:
             $userWithAssignmentsCacheTtl: '%app.cache.ttl.get_user_with_assignments%'
             $ltiVersion: '%app.lti.version%'
             $ltiRegistrationId: '%app.lti1p3.registration_id%'
+            $privateKeyPath: '%app.jwt.private_key_path%'
```
Skyld, line 46: Let's be more specific with the bindings, `$privatekeyPath` can be used in multiple different scenarios. Let's use a more specific name, OR preferably let's autowire it directly on the service which is using it

phpcs.xml
```diff
@@ -15,4 +15,7 @@
     <file>src/</file>
     <file>tests/</file>
 
+    <!--Excluding config/bundles.php to avoid jenkins build failing-->
+    <exclude-pattern>config/bundles\.php</exclude-pattern>
```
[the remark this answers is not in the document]
Skyld: after every bundle install / update this files is regenerated by symfony and if a bundle name is long, code sniffer will complain about too long lines. I think we can have this exclusion to avoid some recurring headaches. 👍

phpstan.neon
```diff
@@ -24,3 +24,9 @@ parameters:
     # Symfony CLI parameter handling
     - message: '#^Cannot cast array<string>(\|bool)?\|string\|null to (int|string).$#'
       path: src/Command
+    # Doctrine dynamic method call error, findOneByXXX methods do not exist, but processed via _call
+    - message: '#^Call to an undefined method OAT\\SimpleRoster\\Repository\\UserRepository\:\:findOneByUsername\(\).$#'
+      path: src/Action/Security/RefreshJWTTokenAction.php
+    # Authentication doesn't seem to be Throwable to phpstan
+    - message: '#^Parameter \#1 \$exception of method OAT\\SimpleRoster\\Responder\\SerializerResponder\:\:createErrorJsonResponse\(\) expects Throwable, Symfony\\Component\\Security\\Core\\Exception\\AuthenticationException\|null given.$#'
```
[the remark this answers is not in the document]
Skyld: Yes, I agree with Andre

src/Service/JWT/JWTManager.php
```diff
@@ -0,0 +1,136 @@
...
+    private $privateKeyPath;
+
+    /** @var string */
+    private $publicKeyPath;
+
+    /** @var string */
+    private $passphrase;
+
+    public function __construct(
+        CacheItemPoolInterface $cache,
```
Skyld, line 59: You forgot to define the cache pool in services.yaml

src/Service/JWT/JWTManager.php
```diff
@@ -0,0 +1,136 @@
...
+use Carbon\Carbon;
+use Lcobucci\JWT\Builder;
+use Lcobucci\JWT\Signer\Key;
+use Lcobucci\JWT\Signer\Rsa\Sha256;
+use Lcobucci\JWT\Token;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class JWTManager
```
Skyld, line 34: Yay.... Manager classes are evil. This is a good example of how to violate SRP. You have 4 public methods where you can:

- Build the JWT token
- Generate cache identifier
- Retrieve a token from cache
- Store a token in cache

These responsibilities should be separated to at least 3 different classes. Please refactor

tests/Traits/UserAuthenticatorTrait.php
```diff
@@ -23,26 +23,26 @@
...
+        $accessToken = $jwt->create($user, $ttl);
 
-        $cookie = new Cookie($session->getName(), $session->getId());
-        $kernelBrowser->getCookieJar()->set($cookie);
+        $refreshToken = $jwt->create($user, $refreshTtl);
+        $jwt->storeTokenInCache($refreshToken, $refreshTtl);
+
+        $kernelBrowser->setServerParameter('HTTP_Authorization', sprintf('Bearer %s', $accessToken));
+
+        return (string)$refreshToken;
```
Skyld, line 46: Whaaaat..? What you should do here is to actually call the login endpoint with the user credentials, grab the token from the response and return it at the end so it can be used for further API calls. Please refactor

Skyld (changes requested):

tests/Traits/UserAuthenticatorTrait.php
uncleempty: Good point. But, it also means (as I wrote to you) that we would be missing possibility to check logs via TestHandler in that case. So, I just did it as it was for normal auth flow -> without sending real request, just creating tokens and setting access one into server params.
Now the question: are you fully sure that it would be better to use "normal" flow with requesting token via API *_for tests_*, instead of just setting it?

docs/devops-documentation.md
```diff
@@ -92,6 +97,20 @@ $ sudo -u www-data bin/console doctrine:schema:update --force
 $ sudo -u www-data bin/console doctrine:schema:drop --force
 ```
 
+- Generate private/public keypair for JWT auth flow to work
+    1. create directory to store keypair
+        ```shell
+        $ mkdir -p config/secrets/prod
+        ```
+    2. generate private key
```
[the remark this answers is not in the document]
uncleempty: Mhm, sorry, not sure what do you mean. Because, actually, when you're generating keys - you would be asked for passphrase

composer.json
```diff
@@ -72,7 +73,8 @@
         "symfony/debug-bundle": "^5.0",
         "symfony/phpunit-bridge": "^5.0",
         "symfony/profiler-pack": "^1.0",
-        "symfony/var-dumper": "^5.0"
+        "symfony/var-dumper": "^5.0",
+        "symfony/web-server-bundle": "4.4"
```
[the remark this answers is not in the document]
uncleempty: > Hmm... are you sure we are not missing some direct dependencies from our composer? Libraries like lcobucci/jwt? We should have them listed here I think

That's a good question: because for now `lcobucci/jwt` is a dependency of `oat-sa/lib-lti1p3-core` which we are dependent from. Since we cannot go with a version higher than the one guys require in lti1p3-core - do you think it would be useful to set version in here, @Skyld ?

docs/devops-documentation.md
```diff
@@ -92,6 +97,20 @@ $ sudo -u www-data bin/console doctrine:schema:update --force
 $ sudo -u www-data bin/console doctrine:schema:drop --force
 ```
 
+- Generate private/public keypair for JWT auth flow to work
```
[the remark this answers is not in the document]
uncleempty: Already added into the docker/local docs

tests/Traits/UserAuthenticatorTrait.php
Skyld: Yes

composer.json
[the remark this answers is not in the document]
Skyld: Yes since multiple features are depending on that library. If we remove the LTI 1.3 bundle one day (theoretically) we don't want to break the JWT authentication.

phpcs.xml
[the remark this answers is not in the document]
uncleempty: As Dani said - the main problem is that this file is re-generated after each install/update. And in case, each env would be specified in options for the bundle - line would easily exceed 120 chars. BUT, this is not any kind of a problem, because file is auto-generated. For that reason I decided to suppress this warning for that file

tests/Traits/UserAuthenticatorTrait.php
Skyld: You don't care about the login logs 99% of the time, the only place you want to test it when you test the login endpoint itself. In other test cases you are not asserting on it

uncleempty: > You don't care about the login logs 99% of the time, the only place you want to test it when you test the login endpoint itself. In other test cases you are not asserting on it
Not login logs. What I meant is - if you remember on prev week we've discussed the problem I had with one test failing, because TestHandler didn't see any log records. As I've found - this is because sending the request to login, as after the request the logger handlers would be cleared. So, yeah, why not I can switch it back to use request for authorization via JWT, BUT, this would mean that I would drop some test assertions checking the log records, as it would be impossible.
So, are we agreed on that?

phpstan.neon
```diff
@@ -24,3 +24,9 @@ parameters:
     # Symfony CLI parameter handling
     - message: '#^Cannot cast array<string>(\|bool)?\|string\|null to (int|string).$#'
       path: src/Command
+    # Doctrine dynamic method call error, findOneByXXX methods do not exist, but processed via _call
```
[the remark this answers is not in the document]
uncleempty: Tried, either I'm cooking it improperly or it doesn't work.
Actually, this way is proposed in phpstan issues comments: https://github.com/phpstan/phpstan-doctrine/issues/14

tests/Traits/UserAuthenticatorTrait.php
Skyld: No we don't. If you authenticate the user with this trait AND set up the test log handler after, you'll be able to assert on logs. Let's refactor, and then I can help you with the log assertions. Here is an example by the way where we assert on logs after authentication: `GetUserAssignmentLtiLinkActionTest->testItLogsSuccessfulLtiRequestCreation()`

docs/devops-documentation.md
```diff
@@ -29,6 +29,11 @@ The main configuration file is `.env`, located in root folder.
 | APP_SECRET | Application secret |
 | APP_API_KEY | Application API Key |
 | APP_ROUTE_PREFIX | Application route prefix, [default: `/api` ]. Details: [Applying custom route prefix](#applying-custom-route-prefix)
+| JWT_SECRET_KEY | Path to RSA private key for JWT Auth flow |
```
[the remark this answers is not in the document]
andreluizmachado: What I'm proposing is to have only one pair of key for the application. This keyes would be used for all of our security issues (lti, jwt, ...). If we agree on it, then we wouldn't need to have different keys every time that we implement something which uses encryption. We could have one central point of keys. we could use something like: 
```shell script
#.env
PASSPHRASE =XXX
PRIVATE_KEY_PATH=
PUBLIC_KEY_PATH=
...
```

```yaml
lti1p3.yaml
...
        platformKey:
            key_set_name: "platformSet"
            public_key: "${PUBLIC_KEY_PATH}"
            private_key: "${PRIVATE_KEY_PATH}"
            passphrase: "${PASSPHRASE}"

...
```

```yaml
config/services.yaml
...
    app.public_key_path: '%env(resolve:PUBLIC_KEY_PATH)%'
    app.private_key_path: '%env(resolve:PRIVATE_KEY_PATH)%'
    app.passphrase: '%env(resolve:PASSPHRASE)%'
...
```

andreluizmachado (changes requested):

config/secrets/test/jwt_private.pem
```diff
@@ -0,0 +1,54 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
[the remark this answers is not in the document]
andreluizmachado: this woud be my idea:
https://github.com/oat-sa/simple-roster/pull/176/files#r540907877

andreluizmachado (changes requested):

config/secrets/test/jwt_private.pem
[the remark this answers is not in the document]
uncleempty: Sorry, but this won't change. Because of the concern that different functionality even that may be using "same" keys, should be separated fully. And, especially, feature of the app vs "feature" of the dependency. I guess, noone wants to solve problems later if for some reason smth would change in lti1p3 library and we would be in situation when we are dependent from them and need to either re-generate keys or whatsoever. Better to have it separated.
p.s.
I'll leave this conversation opened

config/services.yaml
```diff
@@ -20,6 +20,11 @@ parameters:
     app.lti1p1.load_balancing_strategy: '%env(resolve:LTI_INSTANCE_LOAD_BALANCING_STRATEGY)%'
     app.lti1p3.registration_id: '%env(resolve:LTI1P3_REGISTRATION_ID)%'
     app.lti.outcome_xml_namespace: '%env(LTI_OUTCOME_XML_NAMESPACE)%'
+    app.jwt.private_key_path: '%env(resolve:JWT_SECRET_KEY)%'
```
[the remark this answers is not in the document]
uncleempty: Look at my answer above, pls, this one would be "closed"

docs/devops-documentation.md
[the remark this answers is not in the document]
uncleempty: Look at my answer above, pls, this one would be "closed"

src/Entity/User.php
uncleempty: First comment - agree, would be changed.
Second - to be discussed about better place. But the intention is to get the assignments of _current user_ which can be cancelled. In any case we already have a "link" to _all_ assignments of the user inside user object.
Third - yes, checked, for me no additional queries were produced.

config/services.yaml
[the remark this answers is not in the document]
uncleempty: Look at my answer above, pls, this one would be "closed"

src/Action/Security/RefreshJWTTokenAction.php
uncleempty: Refactored

openapi/api_v1.yml
```diff
@@ -78,14 +83,40 @@ paths:
                 500:
                     $ref: '#/components/responses/InternalServerErrorResponse'
 
+    /v1/auth/token/refresh:
```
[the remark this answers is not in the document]
uncleempty: Nothing needed to be informed.
And the application knows only one thing - if the refresh token is valid, it can give you new access token.
That's all

openapi/api_v1.yml
```diff
@@ -238,13 +269,13 @@ paths:
...
     securitySchemes:
-        cookieAuth:
-            type: apiKey
-            in: cookie
-            name: SESSIONID
         apiKeyAuth:
             type: http
             scheme: bearer
+        bearerAuth:
+            type: http
```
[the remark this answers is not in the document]
uncleempty: Nope. it shouldn't be.
https://swagger.io/docs/specification/authentication/bearer-authentication/
According to docs, type is `http`

phpstan.neon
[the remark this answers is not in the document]
uncleempty: Sorry, guys, but I don't see any other way to do this, after searching phpstan docs.
`AuthenticationException` extends `RuntimeException` which extends `\RuntimeException` which extends `Exception` which  is implementing goddamn `Throwable` interface. So, I don't see any way to change this except adding exclusion to phpstan rules.
p.s.
Ofc, this can be workaround with either changing underlying interfaces for Authentication providers (but who wants to put hands into framework core?) or by removing dependency from serializer responder, and manually preparing response somewhere in this class.
====
In any case - sorry, guys, but I'd prefer to leave this as is.
p.s
Conversation would be left opened

luisfn (commented):
Good work

Have just small question/suggestions. Could not find anything aside what the guys mentioned before.

Skyld (approved):

Skyld (approved):

Skyld (approved):

config/secrets/test/jwt_private.pem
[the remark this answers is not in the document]
andreluizmachado: IMO the security keys should be part of the application not part of each feature. So according to what you are saying every time that we create a new feature which requires security layer with key pairs we will need to create new key pairs. For me this I very bad for multiple reasons. Just look our environment vars they're growing a lot because of these kind of decisions. Take a look into other projects on the GitHub for example. You will see that most of the time the key pair belongs to the application not the feature. :(

andreluizmachado (approved):
- [x] New code is not subject to concurrency issues (if applicable)

PR #198, opened 2021-03-02
[AUT-16] Tao webhook - rename deliveryURI to remoteDeliveryId

bartlomiejmarszal:
On AUT-16 development we realised that there is a mismatch between event attribute name. This PR introduce fix for this issue. 

https://oat-sa.atlassian.net/browse/AUT-16

PR #202, opened 2021-03-08
[RO-238] Tao webhook - rename event name value

andreluizmachado:
# Introduction
This PR is intended to complete the issue https://oat-sa.atlassian.net/browse/RO-238.
During the integration tests between all the enviornments (Delivery, Authoring, Simple Roster, Test Taker Portal) we've noticed that the eventName is not the correct one. To make the endpoint work properly, we have to change the eventName to the correct one.

## Fixed
- Renamed webhook event name from `RemoteDeliveryPublicationFinished` to `oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent`.

Refs: #173, #198

## Current Situation
The line items will not be updated because unknow events are ignored

## After this PR
The line items will be updated by using the event name `oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent`

### DIFF
--- a/src/WebHook/UpdateLineItemDto.php
+++ b/src/WebHook/UpdateLineItemDto.php
@@ -31,7 +31,7 @@ class UpdateLineItemDto implements JsonSerializable
     public const STATUS_ERROR = 'error';
     public const STATUS_IGNORED = 'ignored';
 
-    public const NAME = 'RemoteDeliveryPublicationFinished';
+    public const NAME = 'oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent';
 
     /** @var string */
     private $id;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- Renamed webhook event name from `RemoteDeliveryPublicationFinished` to `oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent`.
+
 ## 2.0.5 - 2021-03-03
 
 ### Fixed
--- a/docs/features/update-line-items-webhook.md
+++ b/docs/features/update-line-items-webhook.md
@@ -32,7 +32,7 @@ curl --location --request POST 'http://simple-roster.docker.localhost/api/v1/web
 	"events":[
         {
 			"eventId":"52a3de8dd0f270fd193f9f4bff05232c",
-			"eventName":"RemoteDeliveryPublicationFinished",
+			"eventName":"oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent",
 			"triggeredTimestamp":1565602390,
 			"eventData":{
 				"alias":"line-item-slug",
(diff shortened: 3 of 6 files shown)


src/WebHook/UpdateLineItemDto.php
```diff
@@ -31,7 +31,7 @@ class UpdateLineItemDto implements JsonSerializable
     public const STATUS_ERROR = 'error';
     public const STATUS_IGNORED = 'ignored';
 
-    public const NAME = 'RemoteDeliveryPublicationFinished';
+    public const NAME = 'oat\\taoPublishing\\model\\publishing\\event\\RemoteDeliveryCreatedEvent';
```
Skyld, line 34: I assume this is a class namespace on TAO side. I really don't think that we should rely on it, it's very fragile. What do you think?

Skyld (changes requested):

src/WebHook/UpdateLineItemDto.php
uncleempty: Agree, same question from me, what is the gain to use FQCN, instead of just Classname?

uncleempty (commented):
Please see a comment

src/WebHook/UpdateLineItemDto.php
luisfn: Nothing more to say, I'm with the guys :)

andreluizmachado: Agree. Let's include @bartlomiejmarszal on this discussion and see what we can do.

bartlomiejmarszal: In my opinion we should use FQN in configuration. We should never hardcore external webhook in our code

andreluizmachado: Sorry, but the problem is not a configuration, we would stiil need to have a config in our context relying on TAO context this doesn't solve the problem.

bartlomiejmarszal: The simple-roster is relying on external application that can call your api. 
So we have 2 choices IMO:
1. simple-roster adapts to Tao webhook
2. Tao webhook adapts to simple-roster
Either of choices should include configuration to allow adapt to each other without changes in code like external namespaces. 
This is why my proposal is:
- we should create in your code a service that will accept `webhoook configuration`, that will allow Tao web hooks work in simple roster. 
We encounter already this problem with payload of this request.

andreluizmachado: No Simple Roster is not Relying on any kind of appiication. TAO is relying on the Simple Roster via WebHook.
About the 2 options what we would like you to consider here is to remove the FQDNs. Reason: Everybody that will consume this webhooks will have to know about the TAO context in their domains. Just imagine if you have to do it for external clients. they will know about the TAO event classes. is that what you want?

Again the configuration it won't sovle the problem because somewhere into our application (it doesn`t matter if is code or env var or .yaml) will know the fully qualified domain name of your event. So this definitely doesn't solve the issue.

For me what you have to think is about your webhook interface. Do you really want to expose your FQDNs on your webhook interface? because this is your domain. You should decide what would be the best? We're just a consumer.

bartlomiejmarszal: > Everybody that will consume this webhooks will have to know about the TAO context in their domains. Just imagine if you have to do it for external clients

Yes. This is exactly what I'm expecting. Anybody who consume webhook call should exactly know what he can expect from this call, which event he is subscribing to and what is this call payload. This is why I find extremely important to keep this adaptation in configuration or library.

andreluizmachado: There is no gain to have it as a configuration. One the implementation is done, It will not change, This is not a configuration. This is a constant.

mike-ionut-mihai-sandu-tao: i would provide my the following from my perspective : 
 - move the name to a .env but leave a default 
 - let's not use FQN because you add coupling to the library and we don't want to do that,
 - i would document this webhook / name or even interface expected inside the readme
 because this is not in a authoring ownership,
 the simple roaster team should decide the way of implementing this according to the company standards

andreluizmachado: > Yes. This is exactly what I'm expecting. Anybody who consume webhook call should exactly know what he can expect from this call, which event he is subscribing to and what is this call payload. This is why I find extremely important to keep this adaptation in configuration or library.
Do not get me wrong but could you explain to me what this has to do with being configurable or not? I do not see your point here.

Regarding the library, I believe that it makes sense, but It is out of scope now. Do you plan to create a new library to encapsulate this as TAO is the webHook provider? I do not see this reasonable.

bartlomiejmarszal: If you are going to introduce hardcoded expectations such as event name you are internally coupling with external application behaviour. If you doing this in configuration you are separating your internal business logic from any external coupling.

bartlomiejmarszal: Guys, after team consultation regarding this event we agreed that it's too risky to change name of this event to simple class name. We have to ensure we comply with backwards compatibility so existing consumers of this webhook will not loose functionality.

Skyld (approved):

luisfn (commented):
**Review Checklist**
- [ ] Tests are running successfully (old and new ones) on my local machine (if applicable)
- [ ] New code is respecting best practices
- [ ] New code is not subject to concurrency issues (if applicable)
- [ ] Feature is working correctly on my local machine (if applicable)
- [ ] Acceptance criteria are respected
- [ ] Pull request title and description are meaningful
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

luisfn (approved):

luisfn (approved):

Skyld (approved):

PR #206, opened 2021-03-10
[RO-98] Introduced fixed window rate limiter feature

andreluizmachado:
This PR is intended to implement the issue https://oat-sa.atlassian.net/browse/RO-98.

### Added
- Added the possibility to limit the number of requests per endpoint based on a fixed window.
- Added `RATE_LIMITER_STORAGE_DSN`, `RATE_LIMITER_FIXED_WINDOW_ROUTES`, `RATE_LIMITER_FIXED_WINDOW_LIMIT` and `RATE_LIMITER_FIXED_WINDOW_INTERVAL` environment variables.
- Added `LOCK_DSN` environment variable

## Current Situation
There is no way to limit the number of requests.

## After this PR
We'll possibility to limit the number of requests per endpoint based on a fixed window.


config/services.yaml
```diff
@@ -26,6 +26,7 @@ parameters:
     app.jwt.refresh_token_ttl: '%env(int:JWT_REFRESH_TOKEN_TTL)%'
     app.jwt.cache.host: '%env(resolve:REDIS_JWT_CACHE_HOST)%'
     app.jwt.cache.port: '%env(int:REDIS_JWT_CACHE_PORT)%'
+    app.rate_limiter.fixed_window_routes: '%env(csv:RATE_LIMITER_FIXED_WINDOW_ROUTES)%'
```
luisfn, line 29: why this prefix `csv:`?

.env
```diff
@@ -91,3 +91,22 @@ BLACKFIRE_SERVER_TOKEN=<your_backfire_server_token>
 
 WEBHOOK_BASIC_AUTH_USERNAME=webhook
 WEBHOOK_BASIC_AUTH_PASSWORD=cCVMxdX8hfctDefk
+
+###> symfony/rate-limiter ###
+# Choose one of the storages below
+# RATE_LIMITER_STORAGE_DSN=redis://localhost/1
+# Inform the route names that you want to apply fixed window rate limit separeted by comma
+# RATE_LIMITER_FIXED_WINDOW_ROUTES="getAccessToken,refreshAccessToken,bulkCreateUsersAssignments"
```
luisfn, line 99: Not sure why but testing it with `RATE_LIMITER_FIXED_WINDOW_ROUTES="updateLineItems,getAccessToken,healthCheck,getUserAssignments"
`, only `getAccessToken` was not being restricted. Could you check it?

luisfn (changes requested):
Seems really good, thanks for this PR

Just a few things notice, please check and I think we are good to go

.env
andreluizmachado: Thanks @luisfn. Fixed.

src/EventSubscriber/FixedWindowRateLimiterSubscriber.php
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+        $this->fixedWindowLimiter = $fixedWindowLimiter;
+        $this->fixedWindowRoutes = $fixedWindowRoutes;
+        $this->logger = $securityLogger;
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::REQUEST => ['onKernelRequest', 17],
```
Skyld, line 56: What is the reason of specifying the priority?

tests/Functional/Feature/FixedWindowRateLimiterFeatureTest.php
```diff
@@ -0,0 +1,321 @@
...
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ *  Copyright (c) 2020 (original work) Open Assessment Technologies S.A.
```
Skyld, line 18: 2021

tests/Unit/EventSubscriber/FixedWindowRateLimiterSubscriberTest.php
```diff
@@ -0,0 +1,38 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
```
Skyld, line 18: 2021

src/EventSubscriber/FixedWindowRateLimiterSubscriber.php
```diff
@@ -0,0 +1,90 @@
...
+namespace OAT\SimpleRoster\EventSubscriber;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+
+class FixedWindowRateLimiterSubscriber implements EventSubscriberInterface
```
Skyld, line 32: Let's make this feature togglable by introducing a `IS_RATE_LIMITER_ENABLED` environment variable with default value `false`.

docs/features/fixed-window-rate-limiter.md
```diff
@@ -0,0 +1,24 @@
...
+| `RATE_LIMITER_STORAGE_DSN`           | The data source name of the rate limiter storage. Ex: `redis://localhost`                                                                                           |
+| `RATE_LIMITER_FIXED_WINDOW_ROUTES`   | A list of route names (defined in [routes.yaml](../../config/routes.yaml)) separeted by comma. Ex: `"getAccessToken,refreshAccessToken,bulkCreateUsersAssignments"` |
+| `RATE_LIMITER_FIXED_WINDOW_LIMIT`    | Number of requests that can be done during the given interval                                                                                                       |
+| `RATE_LIMITER_FIXED_WINDOW_INTERVAL` | The value of the interval option must be a number followed by any of the units accepted by the PHP date relative formats (e.g. 3 seconds, 10 hours, 1 day, etc.)    |
+| `LOCK_DSN`                           | The data source name of the LOCK storage. Ex: `redis://localhost`                                                                                                   |
```
Skyld, line 24: It would be nice to add a little bit more explanation to these variables. Why do we need two storages here? What will be stored exactly there?

Also I don't think that it's convenient to specify all the different routes one by one. It's good to have this flexibility, but the default value could be `all` or something like this. What do you think?

.env.docker
```diff
@@ -80,3 +80,15 @@ BLACKFIRE_SERVER_TOKEN=<your_backfire_secret>
 
 WEBHOOK_BASIC_AUTH_USERNAME=webhook
 WEBHOOK_BASIC_AUTH_PASSWORD=cCVMxdX8hfctDefk
+
+RATE_LIMITER_STORAGE_DSN=redis://simple-roster-redis/1
+RATE_LIMITER_FIXED_WINDOW_ROUTES="updateLineItems"
+RATE_LIMITER_FIXED_WINDOW_INTERVAL="2 minutes"
+RATE_LIMITER_FIXED_WINDOW_LIMIT=3
```
Skyld, line 87: I am wondering if need more flexibility here. For now I think it's ok to have one global setting, but later we might need to have a configuration file in place where it is possible to define limits per routes.

Also we'll have to be mindful about the default values on prod, we'll need something like 10 requests / second, to be discussed with the team.

docs/features/fixed-window-rate-limiter.md
Skyld: Also please add these variables to devops documentation as well.

Skyld (changes requested):

.env
luisfn: Reopening this one :)

I've tested tested again and the fix seems good, thanks. Just wondering, how about have a option to apply the rate limiter to `all` routes? Maybe a different option or the use of string all instead of the list. My concern is the size of the API since we have plans to expand it, seems good to have this option. WDYT?

src/EventSubscriber/FixedWindowRateLimiterSubscriber.php
andreluizmachado: The reason is that this event must have more priority than security events (they interrupt the process if the user gets the access token or it is not authenticated for example) and less priority of EventListener\RouterListener because this is the time that the framework sets the attributes on the request (the `_router` attribute).

src/EventSubscriber/FixedWindowRateLimiterSubscriber.php
andreluizmachado: Actually, if you do not specify any route on RATE_LIMITER_FIXED_WINDOW_ROUTES, by default the feature will bbe disabled. Do you see the necessity to add a new env var?

.env.docker
andreluizmachado: I fully agree. If during the usage we see that this configuration is not enough, we can improve it for sure. 
Definitely, we have to discuss these parameters when we go to the production environment with the team :)

docs/features/fixed-window-rate-limiter.md
andreluizmachado: What do you guys think:
```
RATE_LIMITER_FIXED_WINDOW_ROUTES="" //disable the feature
RATE_LIMITER_FIXED_WINDOW_ROUTES="*" //enable for all routes (if we use "all" there is a remote possibility that we may create a route with the same name in the future).

Comment (luisfn) [reply docs/features/fixed-window-rate-limiter.md:24]: `*` is better indeed.

File: src/EventSubscriber/FixedWindowRateLimiterSubscriber.php (hunk 6 above)
Comment (luisfn) [reply src/EventSubscriber/FixedWindowRateLimiterSubscriber.php:32]: IMO only `RATE_LIMITER_FIXED_WINDOW_ROUTES` is enough.

File: tests/Functional/Feature/FixedWindowRateLimiterFeatureTest.php [hunk 9]
```diff
@@ -0,0 +1,282 @@
...
+
+        $this->executeUpdateLineItemsRequest();
+
+        self::assertSame(Response::HTTP_OK, $this->kernelBrowser->getResponse()->getStatusCode());
+
+        $this->executeUpdateLineItemsRequest();
+
+        self::assertSame(Response::HTTP_OK, $this->kernelBrowser->getResponse()->getStatusCode());
+
+        $this->resetKernel();
```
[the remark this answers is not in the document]
uncleempty: > Why do we need to reset kernel between calls?

Since there is more than one http call, calls would be isolated. And for us it means, that logs would be unaccessible after 2nd request (though they do appear in log file).
So, there are two ways, either calling `self::ensureKernelShutdown()` or doing the trick as in [there](https://github.com/oat-sa/simple-roster/blob/develop/tests/Functional/Action/Lti/GetUserAssignmentLtiLinkActionTest.php#L456) (lines 456-458)
p.s.
Though I don't see a real need to do that in additional method

cc @luisfn ^^

uncleempty (changes requested):
Though I'd approve this PR in any other case.
BUT
For now I would strictly recommend to block merge of this until discussions with Ops team (especially @bugalot / @llecaque ). Since SR is usually deployed under CloudFront distribution, which can do this from its side.

Skyld:
Closing this PR based on our internal discussions: Rate limiting will be handled on infrastructure level.

PR #221, opened 2021-05-06
[RO-71] Avoid calling deprecated doctrine ping method

uncleempty:
Ticket: https://oat-sa.atlassian.net/browse/RO-71

### Fixed
- Changed implementation of `HealthCheckService` to avoid calling deprecated `ping` method

### DIFF
--- a/src/Service/HealthCheck/HealthCheckService.php
+++ b/src/Service/HealthCheck/HealthCheckService.php
@@ -61,13 +61,13 @@ class HealthCheckService
         try {
             $testQuery = $connection->getDatabasePlatform()->getDummySelectSQL();
             $testQueryResult = (bool)$connection->fetchOne($testQuery);
-        } catch (Exception $e) {
+        } catch (Exception $exception) {
             $testQueryResult = $connection->isConnected();
 
             $this->logger->critical(
                 sprintf(
                     'DB connection unavailable. Got `%s` exception from DBAL',
-                    $e->getMessage()
+                    $exception->getMessage()
                 )
             );
         }
--- a/src/Service/HealthCheck/HealthCheckService.php
+++ b/src/Service/HealthCheck/HealthCheckService.php
@@ -22,18 +22,24 @@ declare(strict_types=1);
 
 namespace OAT\SimpleRoster\Service\HealthCheck;
 
+use Doctrine\DBAL\Exception;
 use Doctrine\ORM\EntityManagerInterface;
 use OAT\SimpleRoster\Exception\DoctrineResultCacheImplementationNotFoundException;
 use OAT\SimpleRoster\HealthCheck\HealthCheckResult;
+use Psr\Log\LoggerInterface;
 
 class HealthCheckService
 {
     /** @var EntityManagerInterface */
     private $entityManager;
 
-    public function __construct(EntityManagerInterface $entityManager)
+    /** @var LoggerInterface $logger */
+    private $logger;
+
+    public function __construct(EntityManagerInterface $entityManager, LoggerInterface $logger)
     {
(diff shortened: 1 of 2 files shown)


src/Service/HealthCheck/HealthCheckService.php
```diff
@@ -50,8 +56,24 @@ public function getHealthCheckResult(): HealthCheckResult
 
         $cacheStatistics = $resultCacheImplementation->getStats();
 
+        $connection = $this->entityManager->getConnection();
+
+        try {
+            $testQuery = $connection->getDatabasePlatform()->getDummySelectSQL();
+            $testQueryResult = (bool)$connection->fetchOne($testQuery);
+        } catch (Exception $e) {
```
luisfn, line 64: Maybe use `DBALException` or something more specific?

luisfn: Also please rename `$e` for something meaningful

luisfn (changes requested):

src/Service/HealthCheck/HealthCheckService.php
uncleempty: > Maybe use DBALException or something more specific?
That's already a specific exception from DBAL :) It just named like a "core" one. And DBALException itself is deprecated
> Also please rename $e for something meaningful
Will do, thx!

luisfn (approved):
**Review Checklist**
- [X] New code is covered by tests (if applicable)
- [X] Tests are running successfully (old and new ones) on my local machine (if applicable)
- [X] New code is respecting code style rules
- [X] New code is respecting best practices
- [X] New code is not subject to concurrency issues (if applicable)
- [X] Feature is working correctly on my local machine (if applicable)
- [X] Acceptance criteria are respected
- [X] Pull request title and description are meaningful
- [X] Changelog is updated according to changes (if applicable)
- [X] Documentation is updated according to changes (if applicable)

src/Service/HealthCheck/HealthCheckService.php
```diff
@@ -50,8 +56,24 @@ public function getHealthCheckResult(): HealthCheckResult
...
+                )
+            );
+        }
+
         return new HealthCheckResult(
-            $this->entityManager->getConnection()->ping(),
+            $testQueryResult,
             ($cacheStatistics['uptime'] ?? 0) > 0
         );
     }
```
Skyld, line 79: What do you think about this?

```
try {
            $connection = $this->entityManager->getConnection();
            $isDatabaseConnectionAvailable = (bool)$connection->fetchOne(
                $connection->getDatabasePlatform()->getDummySelectSQL()
            );
        } catch (Exception $exception) {
            $this->logger->critical(
                sprintf(
                    'DB connection unavailable. Got `%s` exception from DBAL',
                    $exception->getMessage()
                )
            );

            $isDatabaseConnectionAvailable = false;
        }

        return new HealthCheckResult(
            $isDatabaseConnectionAvailable,
            ($cacheStatistics['uptime'] ?? 0) > 0
        );
```

Skyld (changes requested):

src/Service/HealthCheck/HealthCheckService.php
uncleempty: It can also work. But @Skyld , do you see a big difference and profit in amending (or, ofc, not amending) a call to `$connection->isConnected()`?

Skyld: I just don't get the idea of invoking `isConnected` method in the catch clause. If we have an exception in `SELECT 1` query, the database connection is not ok, so I thought the check could be simplified.

Skyld (approved):

PR #236, opened 2021-09-10
CI pipeline improvements

Skyld:
Purpose of this PR is to slightly improve the performance of our CI pipeline by parallelising the steps within it. Now the pipeline will work like this:

Screenshot 2021-09-10 at 16 13 05


composer.json
```diff
@@ -51,6 +51,12 @@
     "conflict": {
         "symfony/symfony": "*"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/theofidry/AliceBundle"
+        }
+    ],
```
luisfn, line 59: Why this is needed now?

luisfn (commented):
Thanks Dani, really liked how CI/CD is organized now.

My only concern is the time it's taking for the Infection test stage, this is related only to environment performance?

Also, why we need the extra repository now but not before?

composer.json
Skyld: Oh yes, I forgot to explain it in the PR description. It is required due to this unfortunate "incident": https://github.com/nelmio/alice/issues/1089

Skyld:
> Thanks Dani, really liked how CI/CD is organized now.
> 
> My only concern is the time it's taking for the Infection test stage, this is related only to environment performance?
> 
> Also, why we need the extra repository now but not before?

I also noticed the slowliness, I think it is caused by the limitations of the CI server. I agree that we should investigate it together with @serge-salamanka-1pt

luisfn (approved):
Looks good to me!

uncleempty (approved):
:+1: 
LGTM

PR #250, opened 2022-02-07
Feature/lsi 238 store user generation files to s3

jeethufoc:
**Ticket** - https://oat-sa.atlassian.net/browse/LSI-238

**Added** - Store user generation files to S3


SergiiTao (changes requested):

src/Service/Bulk/BulkCreateUsersService.php
```diff
@@ -206,6 +210,8 @@ private function createBulkUserAssignmentData(
             }
         }
 
+        $this->awsS3CsvWriter->writeCsv();
```
SergiiTao, line 213: i think it is better to call in the command class itself. 
it is not related to the creation of students

src/Csv/AwsS3CsvWriter.php
```diff
@@ -0,0 +1,67 @@
...
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Csv;
+
+use Aws\S3\Exception\S3Exception;
+use League\Flysystem\MountManager;
+use League\Flysystem\Filesystem;
+
+class AwsS3CsvWriter
```
SergiiTao, line 29: and lets rename it to FolderSyncService
and configure it an alias in services, so it can be reused with different parameters

config/services.yaml
```diff
@@ -162,3 +162,22 @@ services:
...
+            - '@monolog.logger.user_folder_s3_sync'
+
+    app.s3_folder_sync:
+        alias: OAT\SimpleRoster\Service\AwsS3\FolderSyncService
+
+    Aws\S3\S3Client:
+        arguments:
+            -
+                version: 'latest'
+                region: 'eu-west-1'
```
SergiiTao, line 180: we shouldn't hardcode the region

SergiiTao (changes requested):

src/Service/AwsS3/FolderSyncService.php
```diff
@@ -0,0 +1,84 @@
...
+                        if ($resource !== false) {
+                            $mountManager->putStream($destinationPath, $resource);
+                        }
+                    } else {
+                        $mountManager->copy($sourcePath, $destinationPath);
+                    }
+                }
+            }
+        } catch (S3Exception $exception) {
+            $this->logger->info($exception->getMessage());
```
SergiiTao, line 77: this is an error. so severity should be the error

src/Service/AwsS3/FolderSyncService.php
```diff
@@ -0,0 +1,84 @@
...
+
+            return $exception->getMessage();
```
SergiiTao, line 79: you don't need to return the message. you can throw a new exception right here. better a dedicated one. and catch it later where you want

SergiiTao (changes requested):

config/services.yaml
```diff
@@ -162,3 +162,21 @@ services:
...
+        alias: OAT\SimpleRoster\Service\AwsS3\FolderSyncService
+
+    Aws\S3\S3Client:
+        arguments:
+            -
+                version: 'latest'
+                region: '%env(AWS_S3_ACCESS_REGION)%'
+                credentials:
+                    key: '%env(AWS_S3_ACCESS_KEY)%'
+                    secret: '%env(AWS_S3_ACCESS_SECRET_KEY)%'
```
mianchenia, line 182: Sending AWS key/secret directly is not recommended way to auth. It should be provided by IAM roles inside ec2 instance or ENV variables.
See https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/getting-started_basic-usage.html

src/Service/AwsS3/FolderSyncService.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function copyUserFiles(): ?string
+    {
+        try {
+            $mountManager = new MountManager([
+                self::FILESYSTEM_MOUNT_LOCAL_PREFIX => $this->filesystemLocal,
+                self::FILESYSTEM_MOUNT_S3_PREFIX => $this->filesystemS3,
+            ]);
+
+            $contents = $this->filesystemLocal->listContents(date('Y-m-d'), true);
```
mianchenia, line 60: Path to the folder should be pass as an argument.

config/services.yaml
```diff
@@ -57,6 +63,14 @@ services:
...
             $generatedUsersFilePath: '%app.generated_users_file_path%'
             $projectDir: '%kernel.project_dir%'
+            $awsS3AccessKey: '%app.aws_s3_access_key%'
+            $awsS3AccessSecretKey: '%app.aws_s3_access_secret_key%'
+            $awsS3Version: '%app.aws_s3_version%'
+            $awsS3Region: '%app.aws_s3_region%'
+            $awsS3BucketName: '%app.aws_s3_bucket_name%'
+            $awsS3FilePath: '%app.aws_s3_file_path%'
+            $publicUploadsFilesystem: '@oneup_flysystem.public_uploads_filesystem_filesystem'
+            $localUploadsFilesystem: '@oneup_flysystem.default_uploads_filesystem_filesystem'
```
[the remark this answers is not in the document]
mianchenia: fixed

config/services.yaml
mianchenia: fixed

src/Service/AwsS3/FolderSyncService.php
mianchenia: fixed

PR #254, opened 2022-03-08
fix: change field type from json_array to json

SergiiTao:
json_array type is outdated. changed to json

please run
`comment on column users.roles is '(DC2Type:json)';`
on update if needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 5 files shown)


bugalot (changes requested):
Hello there!

I think we have to make extremely clear that this change requires a manual step. It should be very detailed, in addition with clear signals to Operations that between version X and Y, a manual step is required!

bugalot (approved):

PR #260, opened 2022-03-28
fix: put autogenerated qa and lqa users under configurable sub-folders

SergiiTao:
related to [LSI-618](https://oat-sa.atlassian.net/browse/LSI-618)


src/Service/Bulk/BulkCreateUsersService.php
```diff
@@ -33,70 +35,77 @@
 use OAT\SimpleRoster\Repository\Criteria\FindUserCriteria;
 use OAT\SimpleRoster\Repository\UserRepository;
 use OAT\SimpleRoster\Storage\UserGenerator\StorageInterface;
-use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
```
mianchenia, line 38: Does it use relevant hash algorithm? In other parts of project the deprecated is in use.

mianchenia: Last time I got issues with login logic with other password encoder.

SergiiTao: `Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface` is deprecated
`/api/v1/auth/token` works fine with fresh generated users

vraginya (approved):

src/Service/Bulk/BulkCreateUsersService.php
mianchenia: But what for existed user? Or we don't care?

SergiiTao: have you tried it? :)
I've changed it for encoder only. decoder remains the same, so it obviously should work with existing users.
but yes. it works for both new and old users

mianchenia (approved):

PR #267, opened 2022-05-03
Do not create new QA users if not required

SergiiTao:
Related to [https://oat-sa.atlassian.net/browse/LSI-509](https://oat-sa.atlassian.net/browse/LSI-509)


src/Service/Bulk/BulkCreateUsersService.php
```diff
@@ -105,4 +117,16 @@ public function generate(
             }
         }
     }
+
+    private function filterLineItems(array $lineItems): array
```
vraginya, line 121: In case if function expects a param of array type if would be helpful to add a phpdoc with defining a type of array items
    
```suggestion
    /**
    * @param LineItem[] $lineItems
    */
    private function filterLineItems(array $lineItems): array
```

vraginya (changes requested):
Please see my comments

PR #273, opened 2022-08-16
fix: added parsing of auth header in case LTI data is going there

uncleempty:
Changed the way of getting LTI data out of request when we receive LTI outcome callback.
Added the possibility to parse the Auth header, to get the LTI data out of it, in case it comes inside auth header.

Needed to work with November LTS release of TAO side.


bugalot (approved):
Looks good to me!

bugalot:
Due to the urgency and the surprise of [TR-3785](https://oat-sa.atlassian.net/browse/TR-3785) on TAO DEPP Delivery, we will go like this!

PR #280, opened 2022-11-14
Feature/LSI 1439 SR switch to php8.1

jeethufoc:
Update SR Code to PHP 8.1


src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+use Lcobucci\JWT\Token\Plain;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\Signer\Rsa\Sha256;
+use Lcobucci\JWT\Validation\Constraint\SignedWith;
+use ReflectionClass;
+
+class JwtConfiguration
+{
+    /** @var Configuration object */
+    private object $jwtConfig;
```
mianchenia: Please update type hint.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /** @var InMemory object */
+    private object $jwtInMemoryFilePath;
```
mianchenia: Please update type hint.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+    public function __construct(
+        string $jwtPrivateKeyPath,
+        string $jwtPassphrase,
+        string $jwtPublicKeyPath
+    ) {
+        $this->privateKeyPath = $jwtPrivateKeyPath;
+        $this->passphrase = $jwtPassphrase;
+        $this->publicKeyPath = $jwtPublicKeyPath;
+
+        $this->jwtConfig  = (new ReflectionClass(Configuration::class))->newInstanceWithoutConstructor();
```
mianchenia: I believe it's not the way we want to use jwt lib. It's not safe to create object with bypassing the constructor.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->jwtInMemoryFilePath = (new ReflectionClass(InMemory::class))->newInstanceWithoutConstructor();
```
mianchenia: Please look at the comment above.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function initialise(): Configuration
+    {
+        return $this->jwtConfig->forAsymmetricSigner(
```
mianchenia: This is a static method. So we don't need an instance to call it. Please update the call.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+            new Sha256(),
+            $this->jwtInMemoryFilePath->file($this->privateKeyPath, $this->passphrase),
```
mianchenia: This is a static method. Please update the call.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->jwtInMemoryFilePath->file($this->publicKeyPath, $this->passphrase)
```
mianchenia: This is a static method. Please update the call.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+        );
+    }
+
+    public function parseJwtCredentials(string $credentials): Plain
+    {
+        $parsedToken =  $this->jwtConfig->forUnsecuredSigner()
```
mianchenia: This is a static method. Please update the call.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+            ->parser()->parse($credentials);
+        assert($parsedToken instanceof Token\Plain);
+
+        return $parsedToken;
+    }
+
+    public function verifyJwtKey(): SignedWith
+    {
+        $key = $this->jwtInMemoryFilePath->file($this->publicKeyPath);
```
mianchenia: This is a static method. Please update the call.

src/Security/Authenticator/JwtConfiguration.php
mianchenia: As I see below - we don't really need to create this instances. Please remove them.

src/Action/Security/RefreshAccessTokenAction.php
```diff
@@ -76,10 +76,13 @@ public function __construct(
         $this->accessTokenTtl = $jwtAccessTokenTtl;
     }
 
-    public function __invoke(Request $request, Token $refreshToken): Response
+    public function __invoke(Request $request, Token $refreshToken): JsonResponse
     {
         try {
-            $user = $this->userRepository->findByUsernameWithAssignments($refreshToken->getClaim('aud'));
+            assert($refreshToken instanceof Token\Plain);
```
mianchenia: Assertions are primary made for debug/develop/test when you want `run or die` scenario.  For business logic is much better to use `if/else`, `throw`, etc. Please replace them.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->jwtInMemoryFilePath->file($this->privateKeyPath, $this->passphrase),
+            $this->jwtInMemoryFilePath->file($this->publicKeyPath, $this->passphrase)
+        );
+    }
+
+    public function parseJwtCredentials(string $credentials): Plain
+    {
+        $parsedToken =  $this->jwtConfig->forUnsecuredSigner()
+            ->parser()->parse($credentials);
+        assert($parsedToken instanceof Token\Plain);
```
mianchenia: Assertion. Please replace.

mianchenia (changes requested):
Great job! Please look at my comments. We have to make few updates.

mianchenia:
There is an error in Jenkins pipeline. Please have a look. Maybe we could fix it.

src/Security/Authenticator/JwtConfiguration.php
```diff
@@ -0,0 +1,81 @@
...
+            InMemory::file($this->publicKeyPath, $this->passphrase)
+        );
+    }
+
+    public function parseJwtCredentials(string $credentials): Plain
+    {
+        $parsedToken = Configuration::forUnsecuredSigner()
+            ->parser()->parse($credentials);
+
+        if ($parsedToken instanceof plain) {
```
mianchenia: It should be in upper case.

mianchenia: `Plain`

mianchenia (changes requested):
Good job! Only one little notice.

mianchenia (dismissed):

SergiiTao (approved):

mianchenia (approved):

jeethufoc:
Another release branch created to update SR Code to PHP 8.1

PR #281, opened 2022-12-07
Release/v3.0.0

jeethufoc:
Migration of SR Code Base and Libraries to be PHP 8.1 
Ticket - [LSI-1439](https://oat-sa.atlassian.net/browse/LSI-1439)

[LSI-1439]: https://oat-sa.atlassian.net/browse/LSI-1439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SergiiTao (dismissed):

mianchenia (dismissed):

SergiiTao (dismissed):

SergiiTao (approved):

.env.docker
```diff
@@ -89,6 +89,7 @@ WEBHOOK_BASIC_AUTH_PASSWORD=cCVMxdX8hfctDefk
 GENERATED_USERS_FILE_PATH=var/user-files/
 
 FEATURE_GENERATE_USERS_ON_LINEITEM_UPDATED_ENABLED=1
+FEATURE_GENERATED_USERS_S3_FOLDER_SYNC_ON_LINEITEM_UPDATED_ENABLED=0
```
taorafael, line 92: for consistency, I think this should either be FEATURE_GENERATE_*** or FEATURE_GENERATED_*** for both variables

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.0.0 - 2022-10-30
+### Changed
+- Raised minimum required PHP version from `7.4` to `8.1`.
+- Upgraded lcobucci/jwt library from `3.3` to `4.2.1`
```
taorafael, line 6: My comment then:
> The version 4.2.1 of jwt cannot be used promptly as it breaks several unit tests, we have instead to stick to the version 3.2.5 as the 3.2.6 will prevent itself from working with PHP 8.1 while the 3.2.5 will allow us to work with either version of PHP, being 7.4 or 8.1

If the update to 4.2.1 still have issues you can try rolling it back to the version 3.2.5 as it was the latest 3* version that works on both php 7+ and 8+. I found out the hard way. :)

Had similar issues in the past with the principals portal:
https://github.com/oat-sa/pdf-generator/pull/265

.env.docker
jeethufoc: FEATURE_GENERATE_***  is used to enable whether we needs to generate users during remote publishing webhook
FEATURE_GENERATED_***  is used to to decide whether the generated users needs to save to the S3

So I think we needs to use FEATURE_GENERATE_*** and  FEATURE_GENERATED_** variable name

mianchenia (approved):

tests/Unit/Security/Generator/JwtTokenCacheIdGeneratorTest.php
taorafael, line 1: Very good refactor of the token validation @jeethufoc! Great job!

PR #304, opened 2026-03-19
Feat LSI-7252: consume uploaded events in SR worker, route uploads to PP+SR qu…

arcadiuszpiczak:
https://oat-sa.atlassian.net/browse/LSI-7252

## Summary
This PR introduces the **Simple Roster worker-side messaging setup** for rostering file processing.

It adds SR-side consumption of uploaded-file events and aligns messaging configuration/serialization so SR can reliably receive and process messages from queues.

This PR focuses on worker + messaging wiring only.  

## What Changed
1. Added/updated SR message consumption for `RosteringFileUploadedMessage` in a dedicated worker path.
2. Updated Messenger transport configuration for SR worker queues.
3. Added/updated serializer support needed for queue-based message handling (`encode`/`decode` consistency).
4. Simplified storage contract by removing unused `move()` from file storage API and implementation.
5. Added/updated tests covering:
   - message serializer behavior,
   - upload-service dispatch expectations,
   - SR uploaded-message handler behavior.

## Local Development
1. `simple-roster/localstack/...` is used when running **Simple Roster standalone**.
2. In the full stack, LocalStack is initialized from `docker/localstack/init/...`.

## Verification
1. Relevant SR unit/functional/integration tests were executed successfully.
2. Relevant PP messaging/processor tests were executed successfully.
3. Local queue smoke verification confirms SR/PP workers can consume configured messages.

## New Features
- Introduced automated CSV-based rostering file processing for bulk user and assignment imports
- Generates result files with detailed status tracking and validation error reporting per record


SergiiTao:
@coderabbitai review

src/Repository/AssignmentRepository.php
```diff
@@ -105,4 +106,38 @@ public function findByStateAndUpdatedAtPaged(
...
+
+        $this->getConnection()->executeStatement(
+            'INSERT INTO assignments (user_id, line_item_id, state, attempts_count, updated_at) '
+            . 'VALUES (:userId, :lineItemId, :state, 0, CURRENT_TIMESTAMP)',
+            [
+                'userId' => $userId,
+                'lineItemId' => $lineItemId,
+                'state' => $state,
+            ]
+        );
```
SergiiTao, line 128: should be in a transaction

src/Service/Rostering/RosteringFileProcessor.php
```diff
@@ -0,0 +1,446 @@
...
+        );
+    }
+
+    private function findLineItemIdBySessionName(string $sessionName): ?int
+    {
+        if (array_key_exists($sessionName, $this->lineItemIdsBySessionName)) {
+            return $this->lineItemIdsBySessionName[$sessionName];
+        }
+
+        $lineItemId = $this->lineItemRepository->findIdBySlug($sessionName);
```
SergiiTao: why not use the existing methods?
```
        $lineItemId = $this->lineItemRepository->findOneBy(['slug' => $sessionName]);
```

despite it's slightly more expensive to fetch all fields and hydrate it into the object. later it's easier to validate and manipulate, and more secure if the object is immutable

src/Service/Rostering/RosteringFileProcessor.php
```diff
@@ -0,0 +1,446 @@
...
+            $resultRow[self::RESULT_ERROR_MESSAGE] = $exception->getMessage();
+        } catch (Throwable $exception) {
+            $resultRow[self::RESULT_STATUS] = self::ROW_STATUS_INTERNAL_ERROR;
+            $resultRow[self::RESULT_ERROR_TYPE] = self::ERROR_TYPE;
+            $resultRow[self::RESULT_ERROR_CODE] = self::ERROR_CODE_INTERNAL;
+            $resultRow[self::RESULT_ERROR_MESSAGE] = $exception->getMessage();
+
+            $this->logger->error(
+                sprintf('Unexpected error while importing rostering row: %s', $exception->getMessage()),
+                ['trace' => $exception->getTraceAsString()]
```
SergiiTao: I am not sure we want to see trace in logs. with severity "error"

src/Service/Rostering/RosteringFileProcessor.php
```diff
@@ -0,0 +1,446 @@
...
+        if (array_key_exists($sessionName, $this->lineItemIdsBySessionName)) {
+            return $this->lineItemIdsBySessionName[$sessionName];
+        }
+
+        $lineItemId = $this->lineItemRepository->findIdBySlug($sessionName);
+        if (null === $lineItemId) {
+            $this->lineItemIdsBySessionName[$sessionName] = null;
+
+            return null;
+        }
```
SergiiTao, line 415: In terms of logic, this block is useless. cause next lines do the same. 

but I don't think it's a good idea to cache null, as you don't have invalidation. even though the item is addes it will always be null

SergiiTao (changes requested):

src/Repository/AssignmentRepository.php
arcadiuszpiczak: In this flow replaceForRostering() is already executed inside RosteringFileProcessor::runInTransaction(), so DELETE+INSERT are atomic together with the rest of row processing.

src/Service/Rostering/UserCacheInvalidator.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function invalidateAfterAssignmentChange(string $username): void
+    {
+        $this->invalidate($username, 'assignment');
+    }
+
+    /**
+     * @throws CacheInvalidArgumentException
+     */
+    private function invalidate(string $username, string $trigger): void
```
SergiiTao: can we re-use it here `\OAT\SimpleRoster\EventSubscriber\UserCacheInvalidationSubscriber::clearUserCache` 
cause looks like we've got a code duplication

SergiiTao: and another question. Should we warm it up immediately? like in `\OAT\SimpleRoster\EventSubscriber\UserCacheInvalidationSubscriber::clearUserCache`

arcadiuszpiczak: In the rostering worker we can’t directly reuse the subscriber method, because it is tied to Doctrine lifecycle events (onFlush/postFlush on entities), while the worker performs DBAL updates. We can refactor this into shared logic later, but for now this keeps the worker path simple.

Also, subscriber warm-up currently goes through the cache-warmup transport configured as sync, so warming per imported row would add noticeable overhead. For now I intentionally invalidate cache per row without immediate warm-up. 

I suggest keeping this approach and creating a follow-up ticket to implement asynchronous cache clear + warm-up (preferably batched per import). WDYT?

SergiiTao: I think it's possible to use \OAT\SimpleRoster\Message\WarmUpGroupedUserCacheMessage for that

SergiiTao: ok. please create the ticket

arcadiuszpiczak: I've updated code to WarmUpGroupedUserCacheMessage to warm-up cache

SergiiTao (approved):

PR #305, opened 2026-03-20
Feat LSI-7315: status endpoint

arcadiuszpiczak:
https://oat-sa.atlassian.net/browse/LSI-7252

## Summary
- Added the status endpoint flow for `GET /api/v1/status/{referenceId}`.
- Extracted signed URL generation from status service into a dedicated provider abstraction:
  - `ResultFileUrlProviderInterface`
  - `NullResultFileUrlProvider` (default/non-prod behavior)
  - `S3ResultFileUrlProvider` (prod behavior)
- Moved S3 object key composition to `RosteringFileKeyResolver` (`objectKey()`), removing duplicate logic.
- Added xml migration file
- Added merging file logic + calling PP endpoint

## Summary by CodeRabbit

* **New Features**
  * Added rostering import status endpoint to check processing progress, completion, and any error messages by reference ID
  * Added optional principal portal integration configuration for retrieving and merging rostering status from external systems
  * Enhanced status response with row counters, error details, and generated file URLs

* **Documentation**
  * Updated OpenAPI specification to document the new status endpoint and response format


config/packages/security.yaml
```diff
@@ -26,7 +26,7 @@ security:
             security: false
 
         api_key:
-            pattern: '^%app.route_prefix%/v1/(bulk|line-items|upload)'
```
taorafael, line 29: same concern as before, we are not marking the pattern as the end of the api_key with $, although I see that we might have parameters after, so, yeah, maybe we have to be specific about routes with parameters

config/services_prod.yaml
```diff
@@ -3,3 +3,12 @@ services:
         arguments:
             - version: '%app.aws.version%'
               region: '%app.aws.region%'
+
+    OAT\SimpleRoster\Service\Rostering\ResultFileUrlProviderInterface:
+        alias: OAT\SimpleRoster\Service\Rostering\S3ResultFileUrlProvider
```
taorafael, line 8: same point raised, about using the service as class naming, if not a concern, I'll approve as is

tests/Functional/Action/RosteringImport/GetRosteringImportStatusActionTest.php
```diff
@@ -0,0 +1,160 @@
...
+        $this->kernelBrowser = self::createClient([], ['HTTP_AUTHORIZATION' => 'Bearer testApiKey']);
+        $this->setUpDatabase();
+        $this->fileStorage = self::getContainer()->get(FileStorageInterface::class);
+    }
+
+    public function testItRequiresApiKeyAuthentication(): void
+    {
+        $this->kernelBrowser->request(
+            Request::METHOD_GET,
+            '/api/v1/status/76091d1a-3ef5-438d-a88f-8df73bb5f919',
```
taorafael: maybe this key could be isolated as a variable on the top instead of hardcoded?

taorafael:
the pipeline can't run on it because it's not pointing to develop... I suppose you could change the direction of the PR to develop to trigger the action and then switch back to this queue_worker?

but the PR is still marked as draft

config/services_prod.yaml
arcadiuszpiczak: It seems okay to me - the interface (ResultFileUrlProviderInterface) itself isn’t tied to any specific service naming, while the concrete implementation (S3ResultFileUrlProvider) is already S3-specific, so this wiring feels reasonable.
@taorafael  do you have any suggestions on naming?
@SergiiTao what do you think?

src/Service/Rostering/Dto/RosteringImportStatus.php
```diff
@@ -0,0 +1,151 @@
...
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Service\Rostering\Dto;
+
+use InvalidArgumentException;
+use OAT\SimpleRoster\Entity\RosteringImport;
+
+class RosteringImportStatus
+{
+    private const STATUS_PROCESSED = 'processed';
```
SergiiTao: you can specify the type of constants in 8.4
```suggestion
    private const string STATUS_PROCESSED = 'processed';
```

src/Service/Rostering/Dto/RosteringImportStatus.php
```diff
@@ -0,0 +1,151 @@
...
+
+    public function __construct(
+        private readonly string $referenceId,
+        private readonly string $status,
+        private readonly int $fileLine,
```
SergiiTao, line 17: Is it the count of lines? Could you please name it accordingly?

src/Service/Rostering/S3ResultFileUrlProvider.php
```diff
@@ -0,0 +1,49 @@
...
+                [
+                    'Bucket' => $this->bucket,
+                    'Key' => $objectKey,
+                ]
+            );
+            $request = $this->s3Client->createPresignedRequest($command, $this->signedUrlTtl);
+
+            return (string) $request->getUri();
+        } catch (Throwable $exception) {
+            return null;
```
SergiiTao, line 58: Should we at least log something?

config/services_prod.yaml
SergiiTao: looks OK for me

src/Service/Rostering/RosteringImportStatusService.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function getStatus(string $referenceId): ?RosteringImportStatus
+    {
+        if (!$this->principalPortalEnabled) {
+            $localStatus = $this->resolveLocalStatus($referenceId, true);
+            if ($localStatus === null) {
+                return null;
+            }
+
+            return $localStatus;
```
SergiiTao: should work just like this, no?
```suggestion
            return $this->resolveLocalStatus($referenceId, true);
```

SergiiTao (changes requested):

src/Service/Rostering/S3ResultFileUrlProvider.php
```diff
@@ -0,0 +1,49 @@
...
+
+class S3ResultFileUrlProvider implements ResultFileUrlProviderInterface
+{
+    public function __construct(
+        private readonly FileStorageInterface $fileStorage,
+        private readonly RosteringFileKeyResolver $fileKeyResolver,
+        private readonly S3Client $s3Client,
+        private readonly string $bucket,
+        private readonly string $prefix,
+        private readonly string $signedUrlTtl
```
SergiiTao: Shouldn't it be an int?

SergiiTao:
@coderabbitai review

src/Service/Rostering/S3ResultFileUrlProvider.php
SergiiTao: oh. '+15 minutes' :) ok

src/Service/Rostering/OperationalFileStorage.php
```diff
@@ -16,6 +16,15 @@ public function __construct(
     ) {
     }
 
+    public function exists(string $key): bool
+    {
+        try {
+            return $this->filesystem->fileExists($key);
+        } catch (FilesystemException|Throwable $exception) {
+            return false;
+        }
```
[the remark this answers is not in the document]
SergiiTao: agree. every Exception hear doesn't mean that the file just doesn't exist it can be a i/o, os or hardware problem. So I'd suggest to return false on just FileNotFound exception

src/Service/Rostering/Dto/RosteringImportStatus.php
arcadiuszpiczak: Not really — at least not if we want to match the NG API.

Image

SergiiTao (approved):

taorafael (changes requested):
According to the conversation in the depp-rostering-api group, all mentions to Principal Portal should be generic, without identifying the project.

PR #306, opened 2026-03-24
Feat/lsi 7315/download endpoint

arcadiuszpiczak:
## Summary
Split rostering API responsibilities:
- `GET /api/status/{referenceId}` now returns only import status data.
- Signed download URL is now returned only by `GET /api/download/{referenceId}`.

## Why
Status endpoint should describe processing state only.
Download URL generation is a separate concern and should live in a dedicated endpoint.

## Changes
- Removed signed URL from status payload DTO/serialization.
- Added new download endpoint:
  - `GET /api/download/{referenceId}`
  - response: `{ "signedUrl": "..." }`
- Added API-key security coverage for `/api/download`.
- Added domain exception handling for download URL resolution:
  - service throws `RosteringStatusException`
  - controller maps to `BadRequestHttpException` with consistent message.
- Added/updated unit tests for status and download controllers.

## API Contract
Status response:
```json
{
  "result": {
    "status": "processed",
    "fileLine": 1,
    "messages": []
  }
}

### DIFF
--- a/src/Service/Rostering/Dto/RosteringImportStatus.php
+++ b/src/Service/Rostering/Dto/RosteringImportStatus.php
@@ -15,26 +15,20 @@ class RosteringImportStatus
         private readonly string $referenceId,
         private readonly string $status,
         private readonly int $fileLine,
-        private readonly array $messages,
-        private readonly ?string $resultFileUrl
+        private readonly array $messages
     ) {
     }
 
     public static function pending(string $referenceId): self
     {
-        return new self($referenceId, 'pending', 0, [], null);
+        return new self($referenceId, 'pending', 0, []);
     }
 
     /**
      * @param array<string, mixed> $result
      */
-    public static function fromApiResult(array $result): self
+    public static function fromApiResult(array $result, string $referenceId): self
     {
-        if (!isset($result['referenceId']) || !is_string($result['referenceId']) || trim($result['referenceId']) === '') {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Review by SergiiTao — APPROVED** [review]

File: src/Service/Rostering/Dto/RosteringImportStatus.php [hunk 1]
```diff
@@ -114,38 +93,20 @@ public function getMessages(): array
...
-    }
-
     /**
-     * @return array{referenceId: string, status: string, fileLine: int, messages: array<int, string>, resultFileUrl: ?string}
+     * @return array{status: string, fileLine: int, messages: array<int, string>}
      */
     public function toArray(): array
     {
         return [
-            'referenceId' => $this->referenceId,
```
taorafael, line 144: I actually think the referenceId should be kept, but, its not a blocker

taorafael (approved):
lgtm, very concentrated changes

taorafael (changes requested):
According to the conversation in the depp-rostering-api group, all mentions to Principal Portal should be generic, without identifying the project.

src/Service/Rostering/Dto/RosteringImportStatus.php
arcadiuszpiczak: I thought the same initially - that’s why it was there in the first place.
But since this adapter is meant to match the NG API output, I had to remove it.

PR #302, opened 2026-03-10
Feat: LSI-7314 - upload endpoint oneup flysystem conflict fix

taorafael:
Fix for the PR conflicts on flysystem configuration files.

PR #307, opened 2026-03-24
Revert "Feat: LSI-7314 - upload endpoint oneup flysystem conflict fix"

taorafael:
Reverts *SOME CHANGES* from oat-sa/simple-roster#302

Specifically the ones adding a second local block on top of the oneup_flysystem configuration files

### DIFF
--- a/config/packages/docker/oneup_flysystem.yaml
+++ b/config/packages/docker/oneup_flysystem.yaml
@@ -1,0 +1,21 @@
+# Read the documentation: https://github.com/1up-lab/OneupFlysystemBundle/blob/release/3.x/Resources/doc/index.md
+oneup_flysystem:
+    adapters:
+        user_files_adapter_local:
+            local:
+                location: '%kernel.project_dir%/var/user-files'
+        user_files_adapter_s3:
+            awss3v3:
+                client: Aws\S3\S3Client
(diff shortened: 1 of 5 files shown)


SergiiTao (approved):

config/packages/docker/oneup_flysystem.yaml
```diff
@@ -0,0 +1,21 @@
...
+            local:
+                location: '%kernel.project_dir%/var/user-files'
+        user_files_adapter_s3:
+            awss3v3:
+                client: Aws\S3\S3Client
+                bucket: '%env(FILESYSTEM_OPERATIONAL_S3_BUCKET)%'
+                prefix: '%env(FILESYSTEM_USER_FILES_ADAPTER_S3_PREFIX)%'
+        rostering_files_adapter:
+            local:
+                location: '%kernel.project_dir%/var/rostering-files'
```
arcadiuszpiczak, line 14: docker env should use common volume:

location: '/var/rostering-files'

taorafael: this was just reverting the lines that existed previously, didn't want to put too many changes on top of the revert

kept only the continuous-integration that were making it work and reverted everything from the oneup_flysystem.yaml files

you can make that change on the LSI-7314 as it's the main PR for it all, or I can do it if you don't have the time o>

PR #315, opened 2026-04-20
fix(upload): enforce max records correctly and report actual overflow…

arcadiuszpiczak:
## Summary
This PR fixes CSV max-record validation in upload flow to correctly allow exactly `N` data rows (excluding header) when the limit is set to `N`.

## What was wrong
The previous validation used `$recordCount + 1` in the condition but reported `$recordCount` in the error message.  
This could produce a misleading message like:

`File records count "10000" exceeds maximum allowed records of "10000"`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced CSV file upload validation to accurately report the actual record count when files exceed the configured maximum record limit. Error messages now reflect precise counts for improved troubleshooting.

### DIFF
--- a/src/Service/Upload/UploadedFileValidator.php
+++ b/src/Service/Upload/UploadedFileValidator.php
@@ -90,12 +90,12 @@ class UploadedFileValidator
                     continue;
                 }
 
-                ++$recordCount;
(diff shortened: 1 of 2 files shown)


SergiiTao (approved):

taorafael (approved):
very objective, all solid

PR #316, opened 2026-04-23
feat: allow upload zipped CSVs

SergiiTao:
Allow to upload compressed CSVs as AWS API Gateway has a limit 2Mb per file

## Summary by CodeRabbit

* **New Features**
  * Support uploading a ZIP that contains exactly one CSV; the system extracts and processes the CSV. ZIPs with multiple files are rejected with a clear validation error and no file is stored.

* **Documentation**
  * API spec clarifies accepted upload formats: CSV or a ZIP containing a single CSV.

* **Chores**
  * PHP "zip" extension is now required.

* **Tests**
  * Added unit and functional tests for ZIP extraction, single-file acceptance, multi-file rejection, and related error cases.

### DIFF
--- a/src/Service/Upload/PreparedUploadedFile.php
+++ b/src/Service/Upload/PreparedUploadedFile.php
@@ -10,9 +10,8 @@ final readonly class PreparedUploadedFile
 {
     public function __construct(
         private UploadedFile $file,
-        private bool         $isTemporary = false
-    )
-    {
+        private bool $isTemporary = false
+    ) {
     }
 
     public function file(): UploadedFile
--- a/src/Service/Upload/PreparedUploadedFile.php
+++ b/src/Service/Upload/PreparedUploadedFile.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OAT\SimpleRoster\Service\Upload;
+
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+final readonly class PreparedUploadedFile
+{
+    public function __construct(
+        private UploadedFile $file,
+        private bool         $isTemporary = false
+    )
+    {
+    }
+
+    public function file(): UploadedFile
+    {
+        return $this->file;
+    }
+
+    public function cleanup(): void
+    {
+        if ($this->isTemporary) {
+            @unlink($this->file->getPathname());
+        }
+    }
+}
--- a/src/Service/Upload/UploadFileService.php
+++ b/src/Service/Upload/UploadFileService.php
@@ -71,7 +71,7 @@ class UploadFileService
 
         $archive = new ZipArchive();
 
-        if (!$archive->open($file->getPathname())) {
+        if ($archive->open($file->getPathname()) !== true) {
             throw new UploadedFileValidationException('Unable to read uploaded ZIP file.');
         }
 
--- a/src/Service/Upload/UploadFileService.php
+++ b/src/Service/Upload/UploadFileService.php
@@ -6,13 +6,16 @@ namespace OAT\SimpleRoster\Service\Upload;
 
 use OAT\SimpleRoster\Message\RosteringFileUploadedMessage;
 use OAT\SimpleRoster\Service\Rostering\RosteringFileKeyResolver;
+use OAT\SimpleRoster\Service\Upload\Exception\UploadedFileValidationException;
 use Ramsey\Uuid\Uuid;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\Messenger\MessageBusInterface;
+use ZipArchive;
 
 class UploadFileService
 {
     private const UPLOAD_SUCCESS_MESSAGE = 'File uploaded';
+    private const ZIP_FILE_EXTENSION = 'zip';
 
     public function __construct(
         private readonly UploadedFileValidator $validator,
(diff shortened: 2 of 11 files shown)


src/Service/Upload/UploadFileService.php
```diff
@@ -27,22 +30,86 @@ public function __construct(
...
+
+        if (!class_exists(ZipArchive::class)) {
+            throw new UploadedFileValidationException('ZIP file support is not available.');
+        }
+
+        $archive = new ZipArchive();
+
+        if (!$archive->open($file->getPathname())) {
+            throw new UploadedFileValidationException('Unable to read uploaded ZIP file.');
+        }
```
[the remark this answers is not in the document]
taorafael: @SergiiTao, makes sense this one

src/Service/Upload/UploadFileService.php
```diff
@@ -27,22 +30,86 @@ public function __construct(
...
+            $fileIndex = $fileIndexes[0];
+            $content = $archive->getFromIndex($fileIndex);
+
+            if (!is_string($content)) {
+                throw new UploadedFileValidationException('Unable to extract file from uploaded ZIP archive.');
+            }
+
+            $extractedPath = tempnam(sys_get_temp_dir(), 'upload_zip_');
+            if (!is_string($extractedPath) || file_put_contents($extractedPath, $content) === false) {
+                throw new UploadedFileValidationException('Unable to prepare extracted ZIP file for processing.');
```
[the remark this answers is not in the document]
taorafael: don't think this could be the case with a 2mb limit for the zip file

src/Service/Upload/UploadFileService.php
```diff
@@ -27,22 +30,86 @@ public function __construct(
...
+                'referenceId' => $referenceId,
+            ];
+        } finally {
+            if ($fileToProcess !== $file) {
+                @unlink($fileToProcess->getPathname());
+            }
+        }
+    }
+
+    private function extractFileFromZipIfNeeded(UploadedFile $file): UploadedFile
```
arcadiuszpiczak: I’d consider moving ZIP handling out of UploadFileService. Extracting an archive and creating a temporary UploadedFile feels like a separate responsibility.
A dedicated service would make this easier to test and would keep UploadFileService focused on validation, storing and dispatching the message.

src/Service/Upload/UploadFileService.php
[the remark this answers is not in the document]
arcadiuszpiczak: agreed

ZipArchive::open() should be checked with strict comparison

src/Service/Upload/UploadFileService.php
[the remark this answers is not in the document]
arcadiuszpiczak: I think this is a minor issue.
But we could use ZipArchive::getStreamIndex() and copy it to a temp file with stream_copy_to_stream()

arcadiuszpiczak (changes requested):

src/Service/Upload/UploadFileService.php
```diff
@@ -27,22 +30,86 @@ public function __construct(
...
-        $storageKey = $this->fileKeyResolver->inputFileKey($referenceId);
+        try {
+            $this->validator->validate($fileToProcess);
 
-        $this->fileStorage->store(
-            $file,
-            $storageKey,
-            ['referenceId' => $referenceId]
-        );
+            $referenceId = Uuid::uuid7()->toString();
```
arcadiuszpiczak: Can we consider UUID6 here instead of UUID7?

Reason: PP is still on a legacy stack (PHP 7.4 + older locked UUID libs), and moving it cleanly to library-native UUID7 is currently problematic there.  

If SR moves to UUID7 now, we’ll have UUID version drift between SR and PP until PP is upgraded (unless we want to upgrade PP)

UUID6 would still give us sortable IDs.

arcadiuszpiczak (changes requested):

src/Service/Upload/UploadFileService.php
SergiiTao: just wanted to try, and forgot to roll back :)

src/Service/Upload/UploadFileService.php
SergiiTao: moved out

taorafael (changes requested):
need to update the changelog :)

src/Service/Upload/ZippedCsvExtractor.php
```diff
@@ -0,0 +1,136 @@
...
+
+                return new PreparedUploadedFile(
+                    new UploadedFile($extractedPath, $originalName, null, null, true),
+                    true
+                );
+            } catch (UploadedFileValidationException $exception) {
+                @unlink($extractedPath);
+
+                throw $exception;
+            }
```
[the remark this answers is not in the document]
arcadiuszpiczak: I agree
If another unexpected Throwable happens after tempnam(), the temp file may be left behind.

arcadiuszpiczak (changes requested):

arcadiuszpiczak (approved):

src/Service/Upload/ZippedCsvExtractor.php
```diff
@@ -0,0 +1,136 @@
...
+                $this->extractArchiveEntryToPath($archive, $fileIndex, $extractedPath);
+
+                $name = (string) $archive->getNameIndex($fileIndex);
+                $originalName = basename($name);
+
+                return new PreparedUploadedFile(
+                    new UploadedFile($extractedPath, $originalName, null, null, true),
+                    true
+                );
+            } catch (\Throwable $exception) {
```
arcadiuszpiczak: Approved, just a minor comment: since Throwable is not always an Exception, it might be clearer to name the variable $throwable.

arcadiuszpiczak (approved):

src/Service/Upload/PreparedUploadedFile.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace OAT\SimpleRoster\Service\Upload;
+
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+final readonly class PreparedUploadedFile
+{
+    public function __construct(
+        private UploadedFile $file,
+        private bool         $isTemporary = false
```
taorafael: I don't think it should be aligned like this... in case we add another variable that's longer than $file/UploadedFile it would either force all lines to be updated or make some longer and smoe shorter....

taorafael (approved):
lgtm

PR #332, opened 2026-06-24
Feat/lsi 8130/prepare instructions to install

SergiiTao:


docs/docker-development.md
arcadiuszpiczak: probably `vendor/bin/phpunit` instead of `bin/phpunit`

docs/docker-development.md
```diff
@@ -27,30 +27,36 @@ COMPOSER_AUTH={"github-oauth":{"github.com":"your token here"}}
...
 Then install application dependencies:
 
 ```shell script
-$ docker container exec -it simple-roster-phpfpm composer install
+docker container exec -it simple-roster-phpfpm composer install
+```
+
+Setup the DB
+```shell
+docker container exec simple-roster-phpfpm bin/console doctrine:database:create
```
arcadiuszpiczak: PostgreSQL container already creates the simple-roster database on startup, so this command fails with SQLSTATE[42P04]: Duplicate database. The schema update step is enough after containers are up.

docker-compose-service.yml
```diff
@@ -17,3 +17,5 @@ services:
             - "traefik.http.routers.simple-roster-nginx.rule=Host(`${SR_HOST:-simple-roster.docker.localhost}`)"
             - "traefik.docker.network=oat-docker"
             - "traefik.port=80"
+        ports:
+            - "80:80"
```
arcadiuszpiczak: Image
https://github.com/oat-sa/simple-roster/blob/feat/LSI-8130/prepare-instructions-to-install/docs/docker-development.md

The application is accessible directly via http://localhost/ (port 80), and this setup works as expected. 

However, the access instructions in the documentation are somewhat misleading (screen). They state that the application is "not exposed on any port" and should be accessed via https://simple-roster.docker.localhost/, while the branch actually publishes nginx on host port 80:80. (oat-docker-reverse-proxy from the OAT Docker Stack also conflicts on port 80). Maybe I'm misunderstanding the intended setup.

docker-compose.yml
```diff
@@ -73,17 +73,6 @@ services:
...
-        image: blackfire/blackfire
-        networks:
-            - simple-roster
-        ports:
-            - 8707
-        environment:
-            - BLACKFIRE_SERVER_ID
-            - BLACKFIRE_SERVER_TOKEN
-
     simple-roster-localstack:
```
arcadiuszpiczak, line 76: LocalStack did not work in my clean installation run. The simple-roster-localstack container exited immediately, and the worker then failed because it could not resolve simple-roster-localstack. Could you confirm whether LocalStack is expected to work as part of this setup?

Image

Image

arcadiuszpiczak (changes requested):

docker-compose.yml
SergiiTao: I don't have any problems. May be this is something MacBook-specific?

docker-compose.yml
arcadiuszpiczak: LocalStack is only used internally by the application containers, so we do not need to expose port 4566 on the host. Keeping this port mapping may create conflicts with other local Docker setups without providing any real benefit here

docker-compose.yml
```diff
@@ -73,17 +73,6 @@ services:
...
         container_name: simple-roster-localstack
         image: localstack/localstack
```
arcadiuszpiczak: LocalStack image should be pinned to a specific version instead of using latest. Right now the setup depends on whatever Docker pulls at the moment, which makes the installation flow unstable and non-reproducible
suggested fix: 
`image: localstack/localstack:3.0.2`

docker-compose.yml
arcadiuszpiczak: The LocalStack data volume should use /var/lib/localstack instead of /tmp/localstack. The previous mount caused startup issues on macOS, while /var/lib/localstack matches the expected runtime data directory for current LocalStack versions.
`"./localstack/volume:/var/lib/localstack"`

arcadiuszpiczak (changes requested):

arcadiuszpiczak (approved):

